PR #42, opened 2019-06-28
Added hip-clang support

akilaMD:
This PR adds CI testing for our new hip-clang compiler. For a period of time, we'll have to support both the hcc and the hip-clang compilers, thus we'll have to test both in CI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,7 @@ hipSPARSECI:
     hipsparse.compiler.compiler_path = 'c++'
     
     // Define test architectures, optional rocm version argument is available
-    def nodes = new dockerNodes(['gfx900 && centos7', 'gfx906'], hipsparse)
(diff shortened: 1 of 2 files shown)


ntrost57:
rocSPARSE dependency should not be replaced by repository package in the docker file. This is mainly that we can explicitly select different versions of rocSPARSE for different hipSPARSE builds, e.g. if I want to test cutting edge features in hipSPARSE, I might require latest rocSPARSE that is not yet available on the repository.

akilaMD:
> rocSPARSE dependency should not be replaced by repository package in the docker file. This is mainly that we can explicitly select different versions of rocSPARSE for different hipSPARSE builds, e.g. if I want to test cutting edge features in hipSPARSE, I might require latest rocSPARSE that is not yet available on the repository.

That makes sense to me. Given that the dependency is compiled using hcc instead of hip-clang, I thought that it would be better to use the repository package instead. Let me revert that change.

PR #69, opened 2019-11-15
hipSPARSE changes to support multiple ROCM installation

pruthvistony:
- ROCm stack install path needs to be provided using the
  variable ROCM_PATH as a cmake parameter. If it is not
  provided, ROCm stack components are searched in /opt/rocm
- Package is generated to install into ROCM_PATH by setting
  it to CPACK_PACKAGING_INSTALL_PREFIX, if defined in the env
  otherwise default into /opt/rocm
- RUNPATH is set in the lib only if ROCM_RPATH env is defined
  with the rpath

Signed-off-by: Pruthvi Madugundu <mpruthvi@gmail.com>


pruthvistony:
Updated the changes and removed the usage of CMAKE_INSTALL_PREFIX to search for library path.
Please check the latest commit. 
As soon as I get the CI test results will check the CentOS failures.

PR #86, opened 2020-03-01
Removing ROCm libs from dependency installation list and runpath updates

pruthvistony:


ntrost57:
This should not go into master but develop branch.

ntrost57:
Merged develop into the PR.

PR #89, opened 2020-03-02
Enable support on CentOS-6 

parallelo:
The end goal for these CentOS6-related mods is to support a manylinux2010 build environment on ROCm.  

On CentOS-6, the 'numactl-libs' package does not exist.  The same functionality belongs to the 'numactl' package instead.  Additionally, CentOS-6 doesn't include a '/etc/os-release' file, so we grab the details from the '/etc/centos-release' file instead (if it exists).


parallelo:
Just did a rebase to grab the newest commits on 'develop' that conflicted with this PR.

PR #90, opened 2020-03-03
hipsparseXnnz

YvanMokwinski:
implementation of hipsparseXnnz.

### DIFF
--- a/clients/include/utility.hpp
+++ b/clients/include/utility.hpp
@@ -420,10 +420,10 @@ void gen_dense_random_sparsity_pattern(int m, int n, T* A, int lda, float sparsi
     {
         for(int i = 0; i < m; ++i)
         {
-	  const float d = ((float)rand())/((float)RAND_MAX);	  
-	  A[j * lda + i] = (d < sparsity_ratio)
-	    ? make_DataType<T>(rand())/make_DataType<T>(RAND_MAX)
-	    : make_DataType<T>(0);
+            const float d  = ((float)rand()) / ((float)RAND_MAX);
+            A[j * lda + i] = (d < sparsity_ratio)
+                                 ? make_DataType<T>(rand()) / make_DataType<T>(RAND_MAX)
+                                 : make_DataType<T>(0);
         }
     }
 }
@@ -899,16 +899,16 @@ hipsparseStatus_t host_nnz(hipsparseDirection_t      dirA,
             }
         }
     }
-    
+
     int sum = 0;
 #ifdef _OPENMP
-#pragma omp parallel for reduction (+:sum)
+#pragma omp parallel for reduction(+ : sum)
 #endif
     for(int j = 0; j < mn; ++j)
-      {
-	sum = sum + nnzPerRowColumn[j];
-      }
-    nnzTotalDevHostPtr[0]  = sum;
+    {
+        sum = sum + nnzPerRowColumn[j];
+    }
+    nnzTotalDevHostPtr[0] = sum;
     return HIPSPARSE_STATUS_SUCCESS;
 }
 
--- a/clients/include/utility.hpp
+++ b/clients/include/utility.hpp
@@ -412,6 +412,24 @@ int gen_2d_laplacian(int                  ndim,
 }
 
 /* ============================================================================================ */
+/*! \brief  Generate a random dense matrix */
+template <typename T>
+void gen_dense_matrix(int                  m,
+		      int                  n,
+		      T * A,
+		      int lda)
+{
+
+  for (int j=0;j<n;++j)
+    {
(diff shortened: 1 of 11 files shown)


library/src/hcc_detail/hipsparse.cpp
```diff
@@ -4898,6 +4898,94 @@ hipsparseStatus_t hipsparseZhyb2csr(hipsparseHandle_t         handle,
...
+                                int                       m,
+                                int                       n,
+                                const hipsparseMatDescr_t descrA,
+                                const float*              A,
+                                int                       lda,
+                                int*                      nnzPerRowColumn,
+                                int*                      nnzTotalDevHostPtr)
+{
+    RETURN_IF_ROCSPARSE_ERROR(rocsparse_snnz((rocsparse_handle)handle,
+                                             (rocsparse_direction)dirA,
```
ntrost57: should use a function for conversion (see e.g. how hipsparseOperation_t is converted), else this would result in failure if rocsparse or cusparse change the enums.

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3518,6 +3518,62 @@ hipsparseStatus_t hipsparseZcsric02(hipsparseHandle_t         handle,
...
+                                int                       m,
+                                int                       n,
+                                const hipsparseMatDescr_t descrA,
+                                const float*              A,
+                                int                       lda,
+                                int*                      nnzPerRowColumn,
+                                int*                      nnzTotalDevHostPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseSnnz(
+        (cusparseHandle_t)handle, dirA, m, n, descrA, A, lda, nnzPerRowColumn, nnzTotalDevHostPtr));
```
ntrost57: same here, should use a conversion function for `dirA`

YvanMokwinski: okay

clients/include/utility.hpp
```diff
@@ -411,6 +411,21 @@ int gen_2d_laplacian(int                  ndim,
     return n;
 }
 
+/* ============================================================================================ */
+/*! \brief  Generate a random dense matrix */
+template <typename T>
+void gen_dense_matrix(int m, int n, T* A, int lda)
```
jsandham: Not that it matters here but this generates a random dense matrix containing only 0's and 1' correct? It might be a better interface if you passed in what the min and max values can be for each entry in the matrix. For example:

for(int j = 0; j < n; ++j)
{
    for(int i = 0; i < m; ++i)
    {
        A[j * lda + i] = rand()%(max-min)) + min;
    }
}

This makes it more general and better matches the function name I think? You can have min and max have default parameters still. Just a suggestion though.

clients/include/utility.hpp
```diff
@@ -846,6 +861,49 @@ static inline hipDoubleComplex testing_neg(hipDoubleComplex val)
...
+                           int*                      nnzPerRowColumn,
+                           int*                      nnzTotalDevHostPtr)
+{
+    int mn = (dirA == HIPSPARSE_DIRECTION_ROW) ? m : n;
+    for(int j = 0; j < mn; ++j)
+    {
+        nnzPerRowColumn[j] = 0;
+    }
+
+    for(int j = 0; j < n; ++j)
```
jsandham: Not sure if we use openmp in hipsparse but you could here if that makes sense

clients/include/utility.hpp
YvanMokwinski: I like it.  

What I ultimately want is to control the sparsity %.
Here the idea was, well I want 30% of non-zero coefficients, that's why I roughly wrote rand()%10<7 ? 0 : 1.
Values didn't matter in my case but this is good to pay attention. I want to be able to control the sparsity.

clients/include/utility.hpp
YvanMokwinski: Agree!

library/src/nvcc_detail/hipsparse.cpp
YvanMokwinski: done.

library/src/hcc_detail/hipsparse.cpp
YvanMokwinski: done.

clients/include/utility.hpp
YvanMokwinski: Added some.

clients/include/utility.hpp
YvanMokwinski: Refactored as a dense random sparsity pattern with a sparsity ratio, values are positive and normalized.

PR #91, opened 2020-03-05
Fix hip target to work with hip-clang

amdkila:


amdkila:
Oh, there's a hipSPARSE-internal. This should go there first...I will close it and fix the errors in hipSPARSE-internal

ntrost57:
This will not work as hipSPARSE clients are supposed to be built with standard C++  compiler. It should use hip::host target, however this requires hip::host issue to be fixed (hip::host target sets -h flag which is not expected).

amdkila:
Hmmm ok. I'll keep hip_hcc unless the compiler matches clang, and then use hip_vdi instead?

saadrahim:
> 
> 
> Hmmm ok. I'll keep hip_hcc unless the compiler matches clang, and then use hip_vdi instead?

If there is a problem in the hip compiler layer, it would be better for them to fix it. Do we have a ticket open for this issue?

amdkila:
> > Hmmm ok. I'll keep hip_hcc unless the compiler matches clang, and then use hip_vdi instead?
> 
> If there is a problem in the hip compiler layer, it would be better for them to fix it. Do we have a ticket open for this issue?

yeah, ideally hip::host would work and we should use that.

ntrost57:
There is a ticket open for this issue for quite a while: http://ontrack-internal.amd.com/browse/SWDEV-204247
hipsparse should not link to any of those targets. rocsparse should link to hip::host publicly instead, which is the preferred solution once SWDEV-204247 is in rocm.

PR #96, opened 2020-03-23
Allow choice of install source for rocSPARSE in CI

eidenyoshida:


eidenyoshida:
> Looks good to me. If rocprim and rocsparse are installed in the dockerfile, and installDependencies was enabled, would the rocprim/rocsparse from CI overwrite the packages installed in the dockerfile?

If you look at common.groovy, packages from CI are only installed if project.installLibraryDependenciesFromCI is set to True. project.installLibraryDependenciesFromCI is only set to True by rocJenkins if it detects the variable in the config file.

However if for some reason project.installLibraryDependenciesFromCI was set to true and dockerfile dependencies were installed, I imagine the CI packages would overwrite the repo package if the version number is higher

PR #97, opened 2020-03-26
Denseconversions

YvanMokwinski:
Integrate hipsparseXdense2csr, hipsparseXdense2csc, hipsparseXcsc2dense and hipsparseXcsr2dense
for X=S,D,C,Z.


clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+    {
+        verify_hipsparse_status_success(HIPSPARSE_STATUS_ALLOC_FAILED,
+                                        "!h_nnzPerRowColumn || !d_nnzPerRowColumn || !d_dense_val");
+        return HIPSPARSE_STATUS_ALLOC_FAILED;
+    }
+
+    //
+    // Initialize the entire allocated memory.
+    //
+    for(int i = 0; i < LD; ++i)
```
jsandham: SInce the values are always the same you can just collapse this into one for loop that iterates through the entire array.

clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+                              hipMemcpyDeviceToHost));
+
+    CHECK_HIP_ERROR(hipMemcpy(cpu_csx_col_row_ind.data(),
+                              d_csx_col_row_ind,
+                              sizeof(int) * std::max(nnz, 1),
+                              hipMemcpyDeviceToHost));
+
+    if(argus.unit_check)
+    {
+        for(int i = 0; i < LD; ++i)
```
jsandham: Again collapse into single loop if you want.

clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+        if(DIRA == HIPSPARSE_DIRECTION_ROW)
+        {
+            CHECK_HIPSPARSE_ERROR(csx2dense(handle,
+                                            M,
+                                            N,
+                                            descr,
+                                            (const T*)d_csx_val,
+                                            d_csx_row_col_ptr,
+                                            d_csx_col_row_ind,
+                                            (T*)d_dense_val,
```
jsandham: d_dense_val is already of type T* so I dont think you need the cast

clients/include/testing_dense2csx.hpp
```diff
@@ -0,0 +1,396 @@
...
+    // Testing invalid pointers.
+    //
+    status = dense2csx(handle,
+                       M,
+                       N,
+                       nullptr,
+                       (const T*)d_dense_val,
+                       LD,
+                       d_nnzPerRowColumn,
+                       (T*)d_csx_val,
```
jsandham: d_csx_val is already of type T* so I dont think the cast is needed

clients/include/testing_csx2dense.hpp
jsandham: In fact I think h_dense_val.resize(N*LD, static_cast<T>(-2)); would work

clients/include/testing_dense2csx.hpp
YvanMokwinski: yes, residual from device_vector usage

clients/include/testing_csx2dense.hpp
YvanMokwinski: sure

clients/include/testing_csx2dense.hpp
YvanMokwinski: yes, residual from device_vector usage

clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+    status = csx2dense(handle,
+                       M,
+                       -1,
+                       descr,
+                       (const T*)d_csx_val,
+                       d_csx_row_col_ptr,
+                       d_csx_col_row_ind,
+                       (T*)d_dense_val,
+                       M - 1);
+    verify_hipsparse_status_invalid_pointer(status, "Error: an invalid size must be detected.");
```
jsandham: Should be verify_hipsparse_status_invalid_size?

clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+    status = csx2dense(handle,
+                       M,
+                       -1,
+                       descr,
+                       (const T*)d_csx_val,
+                       d_csx_row_col_ptr,
+                       d_csx_col_row_ind,
+                       (T*)d_dense_val,
+                       LD);
+    verify_hipsparse_status_invalid_pointer(status, "Error: an invalid size must be detected.");
```
jsandham: Should be verify_hipsparse_status_invalid_size?

jsandham: I guess it works here because the pointer one just defaults to invalid value internally but given that we have the verify_hipsparse_status_invalid_size method probably we should use it here?

clients/include/utility.hpp
```diff
@@ -912,6 +912,122 @@ hipsparseStatus_t host_nnz(hipsparseDirection_t      dirA,
     return HIPSPARSE_STATUS_SUCCESS;
 }
 
+template <hipsparseDirection_t DIRA, typename T>
+hipsparseStatus_t host_dense2csx(int                  m,
```
jsandham: This is interesting. I see that you are returning a status. We don't do that in the other host utility calls but maybe we should?

clients/include/testing_dense2csx.hpp
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    if(argus.unit_check)
+    {
+
+        //
+        // Compute the reference host first.
+        //
+
+        host_dense2csx<DIRA, T>(M,
```
jsandham: You write this so that it would return a status but you are not checking it here. Did you want to check it?

clients/include/utility.hpp
```diff
@@ -912,6 +912,122 @@ hipsparseStatus_t host_nnz(hipsparseDirection_t      dirA,
...
+                    *csx_val++         = A[j * ld + i];
+                    *csx_col_row_ind++ = j + base;
+                }
+            }
+        }
+        return HIPSPARSE_STATUS_SUCCESS;
+    }
+    }
+
+    return HIPSPARSE_STATUS_INVALID_VALUE;
```
jsandham: Looking at this though maybe it is unnecessary to return a status. I mean the only what you can get an invalid status is if you used a direction that was not row or column but those are the only options? I am ok with either having it return a status or void so its up to you and Nico

clients/include/testing_csx2dense.hpp
```diff
@@ -0,0 +1,396 @@
...
+                       d_csx_col_row_ind,
+                       (T*)nullptr,
+                       LD);
+    verify_hipsparse_status_invalid_pointer(status, "Error: an invalid pointer must be detected.");
+
+    //
+    // Testing invalid size on M
+    //
+    status = csx2dense(handle,
+                       -1,
```
jsandham: Should be verify_hipsparse_status_invalid_size?

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,246 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      const hipsparseMatDescr_t descr,
+                                      const float*              A,
+                                      int                       ld,
+                                      const int*                nnzPerRow,
+                                      float*                    csrVal,
+                                      int*                      csrRowPtr,
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseSdense2csr(
+        (cusparseHandle_t)handle, m, n, descr, A, lda, nnzPerRow, csrVal, csrRowPtr, csrColInd));
```
jsandham: Do you need a cast here for the matrix descriptor (and for the rest)? i.e. (const cusparseMatDescr_t)descrA?

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,246 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      const hipsparseMatDescr_t      descr,
+                                      const hipsparse_float_complex* A,
+                                      int                            ld,
+                                      const int*                     nnzPerRow,
+                                      hipComplex*                    csrVal,
+                                      int*                           csrRowPtr,
+                                      int*                           csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseCdense2csr(
+        (cusparseHandle_t)handle, m, n, descr, A, lda, nnzPerRow, csrVal, csrRowPtr, csrColInd));
```
jsandham: SImilar here for the complex arrays, i.e. (const cuComplex*)csrVal?

library/include/hipsparse.h
```diff
@@ -1872,6 +1872,195 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
                                 int*                      nnzPerRowColumn,
                                 int*                      nnzTotalDevHostPtr);
 
+/* Description: 
+   This function This function converts the matrix A in dense format into a sparse matrix in CSR format. */
```
ntrost57: `This function This function`

YvanMokwinski: good catch. fixed

clients/include/testing_csx2dense.hpp
YvanMokwinski: True, but question of taste? :) testing only..

clients/include/testing_csx2dense.hpp
YvanMokwinski: good point, removed these unnecessary casts.

clients/include/testing_dense2csx.hpp
YvanMokwinski: removed all the unnecessary casts.

clients/include/testing_csx2dense.hpp
YvanMokwinski: Well, good catch, will fix use the right routine.
Behind the scene, v_h_s_invalid_size and v_h_s_invalid_pointer check for the same hip status invalid value. This is what the hip status proposes. I don't necessarily want to enhance the status right now.

clients/include/testing_csx2dense.hpp
YvanMokwinski: Yes :) fixed.

clients/include/utility.hpp
YvanMokwinski: Maybe we should ... that's a good point. It seems that I didn't follow the convention.
But since I see that I'm not using the returned status, I made the host routines returning void instead.
But maybe we should ...

clients/include/utility.hpp
YvanMokwinski: let's bring this topic to the next meeting.

clients/include/testing_dense2csx.hpp
YvanMokwinski: ;)

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      float*                    csrVal,
+                                      int*                      csrRowPtr,
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseSdense2csr((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           A,
+                                                           lda,
```
ntrost57, line 3647: Should be `ld` instead of `lda`

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      double*                   csrVal,
+                                      int*                      csrRowPtr,
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseDdense2csr((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           A,
+                                                           lda,
```
ntrost57, line 3670: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      hipComplex*               csrVal,
+                                      int*                      csrRowPtr,
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseCdense2csr((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuComplex*)A,
+                                                           lda,
```
ntrost57, line 3693: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      hipDoubleComplex*         csrVal,
+                                      int*                      csrRowPtr,
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseZdense2csr((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuDoubleComplex*)A,
+                                                           lda,
```
ntrost57, line 3716: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      float*                    cscVal,
+                                      int*                      cscRowInd,
+                                      int*                      cscColPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseSdense2csc((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           A,
+                                                           lda,
```
ntrost57, line 3739: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                                           cscRowInd,
+                                                           cscColPtr));
+}
+
+hipsparseStatus_t hipsparseDdense2csc(hipsparseHandle_t         handle,
+                                      int                       m,
+                                      int                       n,
+                                      const hipsparseMatDescr_t descr,
+                                      const double*             A,
+                                      int                       ld,
```
ntrost57, line 3751: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      double*                   cscVal,
+                                      int*                      cscRowInd,
+                                      int*                      cscColPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseDdense2csc((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           A,
+                                                           lda,
```
ntrost57, line 3762: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      hipComplex*               cscVal,
+                                      int*                      cscRowInd,
+                                      int*                      cscColPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseCdense2csc((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuComplex*)A,
+                                                           lda,
```
ntrost57, line 3785: same as above

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      hipDoubleComplex*         cscVal,
+                                      int*                      cscRowInd,
+                                      int*                      cscColPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseZdense2csc((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuDoubleComplex*)A,
+                                                           lda,
```
ntrost57, line 3808: same as above .. I wont mark all other ones ;)

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int*                      cscRowInd,
+                                      int*                      cscColPtr)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseDdense2csc((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           A,
+                                                           lda,
+                                                           nnz__per_columns,
```
ntrost57, line 3763: Should be `nnzPerColumn`

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int                       m,
+                                      int                       n,
+                                      const hipsparseMatDescr_t descr,
+                                      const double*             cscVal,
+                                      const int*                cscRowInd,
+                                      const int*                cscColPtr,
+                                      double*                   A,
+                                      int                       ld)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseScsc2dense((cusparseHandle_t)handle,
```
ntrost57, line 3930: Should be `cusparseDcsc2dense()`

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int                       m,
+                                      int                       n,
+                                      const hipsparseMatDescr_t descr,
+                                      const hipComplex*         cscVal,
+                                      const int*                cscRowInd,
+                                      const int*                cscColPtr,
+                                      hipComplex*               A,
+                                      int                       ld)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseScsc2dense((cusparseHandle_t)handle,
```
ntrost57, line 3951: should be `cusparseCcsc2dense()`

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int                       m,
+                                      int                       n,
+                                      const hipsparseMatDescr_t descr,
+                                      const hipDoubleComplex*   cscVal,
+                                      const int*                cscRowInd,
+                                      const int*                cscColPtr,
+                                      hipDoubleComplex*         A,
+                                      int                       ld)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseScsc2dense((cusparseHandle_t)handle,
```
ntrost57, line 3972: should be `cusparseZcsc2dense()`

YvanMokwinski: shhh :) indeed.

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int*                      csrColInd)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseZdense2csr((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuDoubleComplex*)A,
+                                                           ld,
+                                                           nnzPerRow,
+                                                           (const cuDoubleComplex*)csrVal,
```
ntrost57, line 3718: Shouldn't be const

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -3628,6 +3628,358 @@ hipsparseStatus_t hipsparseZnnz(hipsparseHandle_t         handle,
...
+                                      int                       ld)
+{
+    return hipCUSPARSEStatusToHIPStatus(cusparseCcsc2dense((cusparseHandle_t)handle,
+                                                           m,
+                                                           n,
+                                                           (cusparseMatDescr_t)descr,
+                                                           (const cuComplex*)cscVal,
+                                                           cscRowInd,
+                                                           cscColPtr,
+                                                           (const cuComplex*)A,
```
ntrost57, line 3958: Shouldn't be const

library/src/nvcc_detail/hipsparse.cpp
ntrost57: 'csrVal' cannot be const.

library/src/nvcc_detail/hipsparse.cpp
ntrost57: `A` cannot be const

library/src/nvcc_detail/hipsparse.cpp
YvanMokwinski: indeed.

PR #102, opened 2020-04-02
bsrmv

ntrost57:


clients/tests/test_bsrmv.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <vector>
+
+typedef hipsparseIndexBase_t                                    base;
+typedef hipsparseDirection_t                                    dir;
+typedef std::tuple<int, int, double, double, int, dir, base>    bsrmv_tuple;
+typedef std::tuple<double, double, int, dir, base, std::string> bsrmv_bin_tuple;
+
+int bsr_M_range[]   = {-1, 0, 500, 7111};
+int bsr_N_range[]   = {-3, 0, 842, 4441};
+int bsr_dim_range[] = {1, 3, 5, 9};
```
jsandham, line 40: Maybe add test for negative block dimension

clients/include/testing_bsrmv.hpp
```diff
@@ -0,0 +1,634 @@
...
+                                 descr,
+                                 dval,
+                                 dptr,
+                                 dcol,
+                                 block_dim,
+                                 dx,
+                                 &h_beta,
+                                 dy);
+
+        if(mb < 0 || nb < 0 || block_dim < 0)
```
jsandham, line 376: Should be invalid if block_dim <= 0 I think?

PR #116, opened 2020-06-05
Split dependencies into CentOS 6,7,8 in install script

amdkila:


install.sh
```diff
@@ -159,10 +164,24 @@ install_packages( )
...
+
+        if [[ "${build_clients}" == true ]]; then
+          install_yum_packages "${client_dependencies_centos_8[@]}"
+        fi
+      elif [[ ( "${VERSION_ID}" -ge 6 ) ]]; then
+        install_yum_packages "${library_dependencies_centos_6[@]}"
+
+        if [[ "${build_clients}" == true ]]; then
+          install_yum_packages "${client_dependencies_centos_6[@]}"
+        fi
```
TorreZuk, line 178: 7 will fall into this case, >= is not what you want just -eq , no?

amdkila: Yep, you're definitely right about that, stupid mistake on my part. I've fixed it in my next commit

TorreZuk:
If you want it to be clearer I would suggest you work in reverse chrono order, i.e.
if ( ge 8)
elseif( ge7 )
elseif (ge 6)
else (#%@)
Also didn't sound like 6 was being testd, my hunch for fortran on Centos6 you need the devtoolset-7 version not a base version.  I know this is for Centos8.

amdkila:
> If you want it to be clearer I would suggest you work in reverse chrono order, i.e.
> if ( ge 8)
> elseif( ge7 )
> elseif (ge 6)
> else (#%@)
> Also didn't sound like 6 was being testd, my hunch for fortran on Centos6 you need the devtoolset-7 version not a base version. I know this is for Centos8.

Yeah, that's a good suggestion. The centOS 6 is there due to the name change for the numa library, but that was there before. I could've just checked for centOS 6 and then then appended the centOS 7 install dependency command, but I decided not to

PR #127, opened 2020-07-14
Bsric0

jsandham:
Adding bsric02


library/src/hcc_detail/hipsparse.cpp
```diff
@@ -5490,6 +5500,488 @@ hipsparseStatus_t hipsparseZcsrilu02(hipsparseHandle_t         handle,
...
+
+
+
+
+
+
+
+
+
+
```
ntrost57, line 5513: Can you remove the empty lines?

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -5490,6 +5500,488 @@ hipsparseStatus_t hipsparseZcsrilu02(hipsparseHandle_t         handle,
...
+hipsparseStatus_t
+    hipsparseXbsric02_zeroPivot(hipsparseHandle_t handle, bsric02Info_t info, int* position)
+{
+    // Obtain stream, to explicitly sync (cusparse bsric02_zeropivot is blocking)
```
ntrost57, line 5559: Did you verify that cusparse is actually blocking?

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -5490,6 +5500,488 @@ hipsparseStatus_t hipsparseZcsrilu02(hipsparseHandle_t         handle,
...
+                                             const hipsparseMatDescr_t descrA,
+                                             const float*              bsrValA,
+                                             const int*                bsrRowPtrA,
+                                             const int*                bsrColIndA,
+                                             int                       blockDim,
+                                             bsric02Info_t             info,
+                                             hipsparseSolvePolicy_t    policy,
+                                             void*                     pBuffer)
+{
+    // Obtain stream, to explicitly sync (cusparse bsric02_analysis is blocking)
```
ntrost57, line 5734: Did you verify cusparse is blocking?

library/src/hipsparse_module.f90
```diff
@@ -3153,6 +3170,256 @@ function hipsparseZcsrilu02(handle, m, nnz, descr, csrVal, csrRowPtr, &
...
+                result(c_int) &
+                bind(c, name = 'hipsparseZbsric02_bufferSize')
+            use iso_c_binding
+            implicit none
+            type(c_ptr), value :: handle
+            integer(c_int), value :: dir
+            integer(c_int), value :: mb
+            integer(c_int), value :: nnzb
+            type(c_ptr), intent(in), value :: descr
+            type(c_ptr), intent(in), value :: bsrVal
```
ntrost57, line 3253: bsrVal is non-const (also in other precisions)

library/src/hcc_detail/hipsparse.cpp
jsandham: Their documentation says cusparseXbsric0_zeroPivot is blocking as they call cudaDeviceSynchronize but I did not verify it.

library/src/hcc_detail/hipsparse.cpp
jsandham: Their documentation says it does not support asynchronous execution but I did not verify it. I am actually not sure how I would verify it

library/src/hipsparse_module.f90
jsandham: Are you referring to the fact that in cusparseSbsric0_bufferSize etc pass bsrVAl by non const? I think this is a bug as bsrVAl is not modified by this function. I made the rocsparse routines pass it const.

library/src/hcc_detail/hipsparse.cpp
jsandham: Fixed in latest push

library/src/hipsparse_module.f90
jsandham: Looks like we do the same for csric0 even though cusparse uses non-const

PR #136, opened 2020-09-11
Guard cuda deprecated

jsandham:
- Can now compile hipsparse with CUDA backend (was previously broken)
- When making standalone programs with hipsparse using CUDA backend, deprecation warnings appear if using functionality that is deprecated in CUDA 10 or 11. If compiling with rocsparse backend, no warnings are given.
- If making standalone programs with hipsparse using CUDA backend, then the functionality must not be removed from CUDA runtime version being used. If it has been removed then corresponding hipsparse functionality does not exist and a compile error will be generated.


jsandham:
@ntrost57 I tested this PR using CUDA 11 as that is what is installed on servers. Do you happen to have your own machine that is using CUDA 10? If so can you test this PR to make sure hipSPARSE compiles on a CUDA 10 machine?

library/include/hipsparse.h
```diff
@@ -2318,7 +2426,7 @@ hipsparseStatus_t hipsparseSbsrilu02_analysis(hipsparseHandle_t         handle,
                                               int                       mb,
                                               int                       nnzb,
                                               const hipsparseMatDescr_t descrA,
-                                              const float*              bsrSortedValA,
+                                              float*                    bsrSortedValA,
```
jsandham, line 2429: Note: cusparse does not take a const float*. Therefore hipsparseSbsrilu02 api cannot use a const float* here since we cant cast const float* to float* when using cuda backend.

ntrost57 (approved):
lgtm

PR #140, opened 2020-10-14
hipsparseXgebsr2csr

YvanMokwinski:
Wrapper hipsparseXgebsr2csr


ntrost57 (approved):
lgtm

clients/include/testing_gebsr2csr.hpp
```diff
@@ -0,0 +1,626 @@
...
+            bsr_row_ptr[i + 1] += bsr_row_ptr[i];
+        }
+    }
+
+    m       = mb * row_block_dim;
+    n       = nb * col_block_dim;
+    int nnz = nnzb * row_block_dim * col_block_dim;
+    // Now use the csr matrix as the symbolic for the gebsr matrix.
+    bsr_val.resize(nnz);
+    int idx = 0;
```
jsandham, line 486: Might be good to move this into a utility routine or something so that it can be reused in the other gebsr rountines?

clients/tests/test_gebsr2csr.cpp
```diff
@@ -0,0 +1,215 @@
...
+int gebsr2csr_col_block_dim_range[] = {-1, 0, 1, 2, 4, 8};
+
+hipsparseIndexBase_t gebsr2csr_csr_base_range[] = {HIPSPARSE_INDEX_BASE_ZERO};
+
+hipsparseIndexBase_t gebsr2csr_bsr_base_range[] = {HIPSPARSE_INDEX_BASE_ONE};
+
+hipsparseDirection_t gebsr2csr_dir_range[] = {HIPSPARSE_DIRECTION_ROW, HIPSPARSE_DIRECTION_COLUMN};
+
+// Matrices from files (float and double)
+int gebsr2csr_row_block_dim_range_bin[] = {5};
```
jsandham, line 52: Is it possible to add more tests with different row and col dimensions?

jsandham: I would at least make them different if it takes too long to run with more tests

YvanMokwinski: it is testing with different row and col dimensions, and I also want equal dimensions. That's why I took the same range but reduced.

clients/include/testing_gebsr2csr.hpp
YvanMokwinski: can be done when we will need to reuse. (?)

clients/tests/test_gebsr2csr.cpp
YvanMokwinski: good catch

YvanMokwinski: fixed.

clients/include/testing_gebsr2csr.hpp
YvanMokwinski: next batch.

clients/include/testing_gebsr2csr.hpp
```diff
@@ -0,0 +1,626 @@
...
+        else
+        {
+            verify_hipsparse_status_success(
+                status, "mb >= 0 && nb >= 0 && row_block_dim >= 0 && col_block_dim >= 0");
+        }
+
+        return HIPSPARSE_STATUS_SUCCESS;
+    }
+
+    // Read or construct CSR matrix
```
jsandham, line 420: BSR in comment?

jsandham: Oh this is a CSR matrix but your arrays are bsr_... Thats a bit confusing.

clients/include/testing_gebsr2csr.hpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    if(argus.timing == 1)
+    {
+        filename = argus.filename;
+    }
+
+    int mb = -1;
+    int nb = -1;
+    if(row_block_dim > 0 && col_block_dim > 0)
```
jsandham, line 339: Should this be:

int mb = -1;
int nb = -1;
if(row_block_dim > 0 && col_block_dim > 0)
{
    mb = (m + row_block_dim - 1) / row_block_dim;
    nb = (n + col_block_dim - 1) / col_block_dim;
}

jsandham (changes requested):

clients/include/testing_gebsr2csr.hpp
YvanMokwinski: No, this is to initialize mb and nb to be used to initialize a csr matrix.

clients/include/testing_gebsr2csr.hpp
```diff
@@ -0,0 +1,626 @@
...
+        {
+            fprintf(stderr, "Cannot open [read] %s\n", binfile.c_str());
+            return HIPSPARSE_STATUS_INTERNAL_ERROR;
+        }
+    }
+    else if(argus.laplacian)
+    {
+        mb = nb
+            = gen_2d_laplacian(argus.laplacian, bsr_row_ptr, bsr_col_ind, bsr_val, bsr_idx_base);
+        nnzb = bsr_row_ptr[m];
```
jsandham, line 440: nnzb = bsr_row_ptr[mb]?

jsandham: I would actually use nnz not nnzb as this becomes confusing. I realize you are constructing a csr matrix of size mb by nb instead of m by n so that you can later construct the gebsr matrix by just modifying the values array but I think while the bsr_row_ptr etc arrays are actually representing a CSR matrix, you should use nnz not nnzb.

YvanMokwinski: good catch.

clients/include/testing_gebsr2csr.hpp
```diff
@@ -0,0 +1,626 @@
...
+               != 0)
+            {
+                fprintf(stderr, "Cannot open [read] %s\n", filename.c_str());
+                return HIPSPARSE_STATUS_INTERNAL_ERROR;
+            }
+        }
+        else
+        {
+            double scale = 0.02;
+            if(m > 1000 || n > 1000)
```
jsandham, line 459: if(mb > 1000 || nb > 1000)?

clients/include/testing_gebsr2csr.hpp
jsandham: I realize but its not consistent with what we do in other routines and it is confusing be we usually associate m and n with the number of rows in columns but here it is being used as mb and nb

YvanMokwinski: 'confusing' depends on its definition. I will re-write for you to be less confused.

YvanMokwinski: This is your showstopper?

jsandham: Its not a 'showstopper' but it was something I came across that I think is confusing and so I commented. Here 'm' is used more like 'mb' and then later on down below you write m= mb * row_block_dim; so 'm' is back to being what it actually is. You can achieve what you want and still have mb and m be consistent with what they actually represent and what we do in other files. If you don't want to make the change then that is fine. I just think its confusing and inconsistent so I commented. I am not forcing you to make the change.

YvanMokwinski: You have requested changed. Am good for changes, I will take in consideration your comments and make it more 'consistent' and less 'confusing'. Not a big fan of the word 'confusing' ;) it was usually used by people to say "oh this is complicated", nut here it makes sense in its right definition.

On another note, usually tests are a pain to read. I have some refactoring planned.

clients/include/testing_gebsr2csr.hpp
YvanMokwinski: works in the confusion :)

jsandham (approved):

PR #159, opened 2020-12-16
SpGEMM

ntrost57:


clients/include/hipsparse_test_unique_ptr.hpp
```diff
@@ -257,6 +257,22 @@ namespace hipsparse_test
         }
     };
 
+    struct spgemm_struct
```
YvanMokwinski, line 260: why do we add this layer?

clients/include/utility.hpp
```diff
@@ -4183,22 +4183,22 @@ static void host_csrgeam(int                  M,
...
-                        const I*             csr_row_ptr_B,
-                        const J*             csr_col_ind_B,
-                        const T*             beta,
-                        const I*             csr_row_ptr_D,
-                        const J*             csr_col_ind_D,
-                        I*                   csr_row_ptr_C,
-                        hipsparseIndexBase_t idx_base_A,
-                        hipsparseIndexBase_t idx_base_B,
-                        hipsparseIndexBase_t idx_base_C,
-                        hipsparseIndexBase_t idx_base_D)
```
YvanMokwinski, line 4201: we need to do some cleanup

YvanMokwinski (approved):
LGTM

YvanMokwinski:
some segfaults ...

clients/include/hipsparse_test_unique_ptr.hpp
ntrost57: temporary storage buffer is not propagated to spgemm_copy function, but we need it for rocsparse. We store the pointer there. Users could deallocate it before the call to spgemm_copy but I do not see any other way to stay compatible with cusparse.

ntrost57:
> some segfaults ...

seg faults in the unit tests? Yeah I know, it is because not latest rocsparse package is used. Should be fixed soon.

library/include/hipsparse.h
```diff
@@ -5058,6 +5065,56 @@ hipsparseStatus_t hipsparseSpMV(hipsparseHandle_t           handle,
                                 hipsparseSpMVAlg_t          alg,
                                 void*                       externalBuffer);
 
+/* Description: Compute the sparse matrix sparse matrix product */
+HIPSPARSE_EXPORT
```
jsandham, line 5069: Heads up this function only became available in cuda 11.0. It will fail compilation when using cuda 10.1, 10.2 etc. It needs to be gated with:

#if(!defined(CUDART_VERSION) || CUDART_VERSION >= 11000)
// function
#endif

Similarly in the nvcc hipsparse.cpp file. See my sparseToDense PR.

All of the generic routines I think have this problem. In my sparseToDense hipsparse review I have fixed the other ones I think and I can get that PR to compile on cuda 10.1. Also after getting the compilation to pass I noticed that a bunch of the routines fail the tests but I think that can be fixed in a different task as it happens on a bunch of them.

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -9293,6 +9304,100 @@ hipsparseStatus_t hipsparseSpMV(hipsparseHandle_t           handle,
                                                      externalBuffer));
 }
 
+hipsparseStatus_t hipsparseSpGEMM_createDescr(hipsparseSpGEMMDescr_t* descr)
```
jsandham, line 9307: Need to gate this for cuda 11.0 or greater. See comment above

clients/include/testing_spgemm_csr.hpp
```diff
@@ -0,0 +1,776 @@
...
+                                                       dcsr_col_ind_C,
+                                                       dcsr_val_C,
+                                                       idxType,
+                                                       idxType,
+                                                       idxBaseC,
+                                                       dataType),
+                                    "success");
+
+    // SpGEMM work estimation
+    verify_hipsparse_status_invalid_handle(hipsparseSpGEMM_workEstimation(nullptr,
```
jsandham, line 148: Required cuda 11.0. See comment below.

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -9293,6 +9304,100 @@ hipsparseStatus_t hipsparseSpMV(hipsparseHandle_t           handle,
...
+    return hipCUSPARSEStatusToHIPStatus(
+        cusparseSpGEMM_workEstimation((cusparseHandle_t)handle,
+                                      hipOperationToCudaOperation(opA),
+                                      hipOperationToCudaOperation(opB),
+                                      alpha,
+                                      (cusparseSpMatDescr_t)matA,
+                                      (cusparseSpMatDescr_t)matB,
+                                      beta,
+                                      (cusparseSpMatDescr_t)matC,
+                                      computeType,
```
jsandham, line 9340: Probably should use hipDataTypeToCudaDataType(computeType) in case the hipDataType and cudaDataType dont match in future.

ntrost57:
Agree with the guards for CUDA. Let's move this into a separate PR.

jsandham:
> Agree with the gates for CUDA. Let's move this into a separate PR.

Ok sounds good.

PR #165, opened 2021-01-20
csru2csr and csr2csru added

ntrost57:


library/src/hcc_detail/hipsparse.cpp
```diff
@@ -10609,6 +10657,854 @@ hipsparseStatus_t hipsparseZgebsr2gebsr(hipsparseHandle_t         handle,
...
+    // Quick return
+    if(m == 0 || n == 0 || nnz == 0)
+    {
+        // nnz must be 0 and pBufferSizeInBytes must be valid
+        if(nnz != 0 || pBufferSizeInBytes == nullptr)
+        {
+            return HIPSPARSE_STATUS_INVALID_VALUE;
+        }
+
+        *pBufferSizeInBytes = 4;
```
jsandham, line 10691: pBufferSizeInBytes could still be nullptr here.

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -10609,6 +10657,854 @@ hipsparseStatus_t hipsparseZgebsr2gebsr(hipsparseHandle_t         handle,
...
```
jsandham, line 10745: Could still be nullptr here

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -10609,6 +10657,854 @@ hipsparseStatus_t hipsparseZgebsr2gebsr(hipsparseHandle_t         handle,
...
```
jsandham, line 10853: Could still be nullptr here

library/src/hcc_detail/hipsparse.cpp
jsandham: Oh sorry no I read that wrong. Its an || "or".

jsandham:
Which version of cuda did you check with?

jsandham (approved):

PR #167, opened 2021-02-03
SpMM CSR

jsandham:
Adding SpMM for CSR format to hipSPARSE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [(Unreleased) hipSPARSE 1.10.4 for ROCm 4.1.0]
 ### Added
-- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SparseToDense, DenseToSparse and SpGEMM
+- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SpMM, SparseToDense, DenseToSparse and SpGEMM
 
 ## [hipSPARSE 1.9.6 for ROCm 4.0.0]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Change Log for hipSPARSE
 
+## [(Unreleased) hipSPARSE 1.10.4 for ROCm 4.2.0]
+### Added
+- Generic API support, including SpMM
+
 ## [(Unreleased) hipSPARSE 1.10.4 for ROCm 4.1.0]
 ### Added
-- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SpMM, SparseToDense, DenseToSparse and SpGEMM
+- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SparseToDense, DenseToSparse and SpGEMM
 
 ## [hipSPARSE 1.9.6 for ROCm 4.0.0]
 ### Added
--- a/clients/include/testing_spmm_csr.hpp
+++ b/clients/include/testing_spmm_csr.hpp
@@ -173,8 +173,6 @@ hipsparseStatus_t testing_spmm_csr()
     hipsparseSpMMAlg_t   alg      = HIPSPARSE_SPMM_CSR_ALG1;
     hipsparseStatus_t    status;
 
-    //std::cout << "alg: " << static_cast<int>(HIPSPARSE_SPMM_CSR_ALG1) << std::endl;
-
     // Determine absolute path of test matrix
 
     // Get current executables absolute path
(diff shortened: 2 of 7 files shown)


library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -9341,6 +9341,60 @@ cusparseSpMVAlg_t hipSpMVAlgToCudaSpMVAlg(hipsparseSpMVAlg_t alg)
...
+        return CUSPARSE_SPMM_COO_ALG2;
+    case HIPSPARSE_SPMM_COO_ALG3:
+        return CUSPARSE_SPMM_COO_ALG3;
+    case HIPSPARSE_SPMM_COO_ALG4:
+        return CUSPARSE_SPMM_COO_ALG4;
+    case HIPSPARSE_SPMM_CSR_ALG1:
+        return CUSPARSE_SPMM_CSR_ALG1;
+    case HIPSPARSE_SPMM_CSR_ALG2:
+        return CUSPARSE_SPMM_CSR_ALG2;
+    default:
```
YvanMokwinski, line 9373: I would remove the default case and put the throw after the switch. In this way, if a case is not specified explicitly in the switch then the compiler will generate errors. Otherwise, the default case would handle it and we would know it at runtime only.

YvanMokwinski (approved):
LGTM (detail: see comment)

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [(Unreleased) hipSPARSE 1.10.4 for ROCm 4.1.0]
 ### Added
-- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SparseToDense, DenseToSparse and SpGEMM
+- Generic API support, including Axpby, Gather, Scatter, Rot, SpVV, SpMV, SpMM, SparseToDense, DenseToSparse and SpGEMM
```
ntrost57, line 5: I think thats going to be part of 4.2. 4.1 branching already happened a while ago.

clients/include/testing_spmm_csr.hpp
```diff
@@ -0,0 +1,370 @@
+/* ************************************************************************
+ * Copyright (c) 2018-2019 Advanced Micro Devices, Inc.
```
ntrost57, line 2: 2021

ntrost57 (approved):

clients/include/testing_spmm_csr.hpp
jsandham: Fixed

CHANGELOG.md
jsandham: Fixed

jsandham: I added the "unreleased" lines for rocm 4.2 with spmm. This can be updated with whatever other features in addition to spmm end up going in 4.2 and the hipsparse version can be updated by whoever ends up push hipsparse to master.

library/src/nvcc_detail/hipsparse.cpp
jsandham: Wouldn't that only help us if we did something like: hipSpMMAlgToHCCSpMMAlg(CUSPARSE_SPMM_CSR_ALG1)?
In reality we only ever call this like hipSpMMAlgToHCCSpMMAlg(alg) where alg is not known at runtime.

YvanMokwinski: It's an enum. When you switch an enum, the default case override any missing case.
Removing the default case will generate an error when a new field is added to the enum but is not handled explicitly by the switch. This way, if you touch the enum you will know at compile time everywhere you haven't specified the new case(s). Moreover, might be useful to use the constexpr for this routine since we could expect values known at compile time. But here if I add a new field to the enum, it will compile but tests will fail with the "Non existant .." exception. Compile time vs run time.

jsandham: Ok thanks. It looks like we do this a lot. Ill create a task to remove all these default cases in hipsparse.

jsandham: That is really smart of the compiler to omit an error if I add to the enum but dont have a default case. Like if I read alg from a file at runtime and was calling hipSpMMAlgToCudaSpMMAlg(alg) then the compiler cant know if my case is covered or not. So if I am understanding this correctly, a switch statement that that switches on a enum must cover all cases of the enum or a compile time error is generated?

jsandham: Ok very interesting. Just did a test and you are correct. I had no idea that is how it c++ works. It is able to detect that you may hit the throw and so gives a compile time error. Very cool

YvanMokwinski: No exactly, it will compile ... except if you activate -Wswitch, which throws warning on switch statements if all cases are not covered and the default statement is missing. -Wswitch is activated with -Wall and warnings can be transformed into errors with -Werror. 
So with -Wall -Werror, and if you remove the default statement, you'll have the compiler telling you if your switch is clean or not.
Possible valid cases are covered through the enumeration of all the cases (which is what we do all the time), the compiler helps with checking if you did, and the exception should be thrown outside the switch.

PR #176, opened 2021-04-19
Switch hipSPARSE CI dependency to internal rocSPARSE

eidenyoshida:


ntrost57:
Why do we change the dependency? rocsparse-internal should be a mirror of rocsparse public, and additionally, rocsparse-internal does not have a master branch.

PR #180, opened 2021-05-07
fixing bug with cuda compiler

YvanMokwinski:
use of make_DataType rather than static_cast


jsandham (approved):

saadrahim:
I enable CI for this, please wait for it to pass.

PR #185, opened 2021-06-02
bsrxmv

YvanMokwinski:
add bsrxmv


cgmb:
The most recent build with `ci:no-ccache` enabled seems to have passed while the first failed, but there were other things that changed between the two builds. The two builds used different versions of rocSPARSE (e.g. the passing Ubuntu build installed `rocsparse_1.20.2-e4fd6d6_amd64.deb` and the failing build installed `rocsparse_1.20.2-10974b7_amd64.deb`).

ntrost57 (approved):

clients/include/testing_bsrxmv.hpp
```diff
@@ -0,0 +1,540 @@
...
+#include "utility.hpp"
+
+#include <cmath>
+#include <hipsparse.h>
+#include <string>
+
+using namespace hipsparse;
+using namespace hipsparse_test;
+
+void testing_bsrxmv_bad_arg(void)
```
jsandham, line 40: Should be template<typename T> correct? I think that is what we usually do instead of just testing float.

clients/include/testing_bsrxmv.hpp
```diff
@@ -0,0 +1,540 @@
...
+    int* dbsr_mask_ptr = (int*)dbsr_mask_ptr_managed.get();
+    int* dbsr_row_ptr  = (int*)dbsr_row_ptr_managed.get();
+    int* dbsr_end_ptr  = (int*)dbsr_end_ptr_managed.get();
+    int* dbsr_col_ind  = (int*)dbsr_col_ind_managed.get();
+    T*   dx            = (T*)dx_managed.get();
+    T*   dy            = (T*)dy_managed.get();
+
+    hbsr_mask_ptr[0] = 2;
+
+    hbsr_row_ptr[0] = 1;
```
jsandham, line 435: I think you can just do:
std::vector<int> hbsr_row_ptr = {1, 4};

clients/tests/test_bsrxmv.cpp
```diff
@@ -0,0 +1,221 @@
...
+    EXPECT_EQ(status, HIPSPARSE_STATUS_SUCCESS);
+}
+
+TEST(bsrxmv, bsrxmv_hipComplex)
+{
+    hipsparseStatus_t status = testing_bsrxmv<hipComplex>();
+    EXPECT_EQ(status, HIPSPARSE_STATUS_SUCCESS);
+}
+
+#if 0
```
jsandham, line 51: What is this for?

jsandham (approved):

clients/include/testing_bsrxmv.hpp
ntrost57: Shouldn't matter for bad args in hipsparse, float is fine.

clients/tests/test_bsrxmv.cpp
YvanMokwinski: garbage. Thanks.

clients/include/testing_bsrxmv.hpp
YvanMokwinski: We skip the other types in 33 tests but test all types in the 48 other tests.

clients/include/testing_bsrxmv.hpp
YvanMokwinski: I can do that

clients/include/testing_bsrxmv.hpp
jsandham: Ok if we skip them often then its fine here

clients/include/testing_bsrxmv.hpp
YvanMokwinski: fixed.

clients/include/testing_bsrxmv.hpp
YvanMokwinski: actually much more often since templated bad_args are very often called with <float> only.
I addressed your comment and we call bad arg for the 4 types.

YvanMokwinski: fixed

PR #192, opened 2021-06-18
Disable CCache on CI; Enable devtoolset-7 in Centos 7 in CI

saadrahim:
Install.sh was not honouring CXX set in an environment variable prior to calling the script. This fixes that and also includes a syntax fix for reference to $PATH in a groovy script.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -21,7 +21,7 @@ def runCompileCommand(platform, project, boolean sameOrg=false)
                 set -x
                 cd ${project.paths.project_build_prefix}
                 ${getDependenciesCommand}
-                export PATH=/opt/rocm/hsa/include:$PATH
+                export PATH=/opt/rocm/hsa/include:\$PATH
                 CXX=/opt/rh/devtoolset-7/root/usr/bin/c++ ${project.paths.build_command}
             """
     }
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -14,26 +14,16 @@ def runCompileCommand(platform, project, boolean sameOrg=false)
             getDependenciesCommand += auxiliary.getLibrary(libraryName, platform.jenkinsLabel, 'develop', sameOrg)
         }
     }
-
-    if(platform.jenkinsLabel.contains('centos7'))
-    {
-        command = """#!/usr/bin/env bash
+    String centos7 = platform.jenkinsLabel.contains('centos7') ? 'source scl_source enable devtoolset-7' : ':'
+    
+    command = """#!/usr/bin/env bash
                 set -x
-                cd ${project.paths.project_build_prefix}
-                ${getDependenciesCommand}
-                export PATH=/opt/rocm/hsa/include:\$PATH
-                CXX=/opt/rh/devtoolset-7/root/usr/bin/c++ ${project.paths.build_command}
-            """
-    }
-    else
-    {
-        command = """#!/usr/bin/env bash
-                set -x
+                ${centos7}
                 cd ${project.paths.project_build_prefix}
                 ${getDependenciesCommand}
                 CXX=${project.compiler.compiler_path} ${project.paths.build_command}
             """
-    }
+ 
     platform.runCommand(this, command)
 }
 
--- a/.jenkins/extended.groovy
+++ b/.jenkins/extended.groovy
@@ -19,6 +19,7 @@ def runCI =
     prj.compiler.compiler_name = 'c++'
     prj.compiler.compiler_path = 'c++'
     prj.libraryDependencies = ['rocSPARSE', 'rocPRIM']
+    prj.defaults.ccache = false
 
     // Define test architectures, optional rocm version argument is available
     def nodes = new dockerNodes(nodeDetails, jobName, prj)
--- a/.jenkins/precheckin-cuda.groovy
+++ b/.jenkins/precheckin-cuda.groovy
@@ -18,6 +18,7 @@ def runCI =
     prj.paths.build_command = './install.sh -c --cuda'
     prj.compiler.compiler_name = 'c++'
     prj.compiler.compiler_path = 'c++'
+    prj.defaults.ccache = false
 
     // Define test architectures, optional rocm version argument is available
     def nodes = new dockerNodes(nodeDetails, jobName, prj)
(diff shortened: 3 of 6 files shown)


.jenkins/common.groovy
```diff
@@ -21,7 +21,7 @@ def runCompileCommand(platform, project, boolean sameOrg=false)
                 set -x
                 cd ${project.paths.project_build_prefix}
                 ${getDependenciesCommand}
-                export PATH=/opt/rocm/hsa/include:$PATH
+                export PATH=/opt/rocm/hsa/include:\$PATH
```
YvanMokwinski, line 24: $$ doesn't work?

YvanMokwinski (approved):
if $$ works I would prefer $$

eidenyoshida (approved):

YvanMokwinski:
@saadrahim A description would have been useful.

saadrahim:
Also, I am very certain the CCache is causing build issues on hipSPARSE. @cgmb we should discuss.

cgmb:
> I am very certain the CCache is causing build issues on hipSPARSE

Could you point me at an example?

> Install.sh was not honouring CXX set in an environment variable prior to calling the script.

I don't know of any `install.sh` scripts that do.

eidenyoshida:
The alternative to modifying install.sh is to make sure devtoolset7 is in the path like we do for other libraries like rocBLAS
https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/dc4997d055b0babe963ab142303b0f26a40ad1fb/.jenkins/common.groovy#L8

saadrahim:
> 
> 
> > I am very certain the CCache is causing build issues on hipSPARSE
> 
> Could you point me at an example?
> 
> > Install.sh was not honouring CXX set in an environment variable prior to calling the script.
> 
> I don't know of any `install.sh` scripts that do.

Good point. Perhaps this solution is not appropriate. Forcing devtoolset-7 is likely a better solution.

cgmb:
I think the problem is that the dependency list for the sample executable is incomplete.

I assume that `hipsparse.mod` is created when `hipsparse.f` is compiled to create the `hipsparse_fortran` object library? If so, the Fortran samples probably need to add `$<TARGET_OBJECTS:hipsparse_fortran>` to their `add_executable` call in order to ensure that the samples are always built _after_ `hipsparse.f`.

ccache doesn't support Fortran and is not used when compiling Fortran sources, so when ccache is enabled, the C and C++ compilation gets really, really fast and the Fortran compilation remains (relatively) slow. That allows the C++ library to finish building before the Fortran library, and the samples begin compiling before `hipsparse.mod` has been created, causing the build failure.

YvanMokwinski:
@cgmb add_dependencies(hipsparse hipsparse_fortran) ?

YvanMokwinski:
when compiling  samples we use the same mechanism for both cpp and Fortran files, thus Fortran samples are using roc::hipsparse. Either we make 'roc::hipsparse' waiting for everything to finish on the library side or we need to 'separate' the compilation of the library side in a better way to be able to use a 'roc::hipsparse_fortran' ? What do you think?

cgmb:
> when compiling samples we use the same mechanism for both cpp and Fortran files, thus Fortran samples are using roc::hipsparse

`hipsparse_fortran` doesn't don't have any outputs that end up in the `hipsparse-*.deb` package. We ship the `hipsparse.f` input file and users need to configure their build system to generate `hipsparse.mod` themselves. So for users following along with the samples, `roc::hipsparse` is never going to provide `hipsparse.mod`, no matter how long they wait.

The samples should do things the same way that an end user would have to do things, which I think is just adding `hipsparse.f` to their library or executable's sources. (The CMake OBJECT library is an optimization of that process when you're adding it to multiple executables, as it builds the file once rather than building the file once for each sample executable.)

Now, I'm not very familiar with Fortran, so it may be that there's better ways of shipping our code, but the way of fixing this that would be consistent with the that users of existing hipSPARSE use the Fortran interface would be to create `hipsparse_fortran` in the clients, rather than the library, and have the samples explicitly include it as a source. (rocBLAS, hipBLAS, and hipSOLVER work that way, too.)

cgmb:
(It may be that we're not providing a great user experience for our Fortran users. I honestly wouldn't know. It might be a great idea to export a `roc::hipsparse_fortran` library, but I think that's a separate problem to solve than fixing the build. Making user-visible changes would require a lot more research to be sure we're doing the right thing.)

ntrost57 (approved):

jsandham (approved):

saadrahim:
fortran wrappers are handled in hipfort. We may be duplicating things.

YvanMokwinski:
yes that was my hunch too, is all of this worth to be taken care of?

TorreZuk:
I'm fine with turning off the ccache as doubt it saves much time here anyway, but expect you should try the fix Cory suggests in the near future or this could pop up as another obscure bug until the time we drop fortran API (Hipfort is not the only layer of fortran support for now).   @daineAMD is more the expert on this design but the fortran module compilation and dependency can be pushed to the clients so that you can drop any fortran usage on the library (cleaner design).  The module is compiled by the client as a user may have a better fortran and so wants to make the module for themselves, thus the client building design matches other client usage.   Probably only a couple lines of Cmake need to change to match the hipBLAS pattern and so the compilation order should be set by cmake regardless of ccache (cmake shouldn't change dependency ordering), but I'm not offering to make the changes so you can ignore me :-)   Daine might be willing?

YvanMokwinski:
I'm willing to take it. Will create a JIRA ticket for it.

saadrahim:
I abandoned the install.sh changes as that was inconsistent from other install.sh scripts as @cgmb suggested. This avoids the cmake race condition and is a fix for devtoolset in Centos 7.

YvanMokwinski:
the fix for devtoolset in Centos 7 solves SWDEV-291503, without it the used c++ compiler was the /usr/bin/c++ version 4.8 where __has_attribute is not supported.

.jenkins/common.groovy
cgmb: I don't think $$ works in this kind of string. https://groovy-lang.org/syntax.html#_string_summary_table

YvanMokwinski: would have been nice ...

PR #194, opened 2021-06-29
Add basic rtest.xml

lawruble13:
Add rtest.xml with basic sets. Required by devops.

### DIFF
--- a/rtest.xml
+++ b/rtest.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE testset SYSTEM "/usr/local/share/rtest.dtd">
+<testset>
+    <fileversion>0.1</fileversion>
+    <var name="GTEST_FILTER"></var>
+    <test sets="osdb,psdb">
+        <always>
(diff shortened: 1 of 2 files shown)


YvanMokwinski (approved):

ntrost57 (approved):

rtest.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE testset SYSTEM "/usr/local/share/rtest.dtd">
```
saadrahim, line 2: Is this hardcoding correct?

TorreZuk: That doesn't look windows friendly.   
You could also leave the dtd specification out of the rtest.xml for now and use a dtd for only an external validation process.

lawruble13: For the moment, the DTD isn't used for validation of `rtest.xml`, and this doctype declaration was primarily used for my IDE and can safely be removed. Eventually we'd like to have the dtd at a static location and reference using a public doctype.

lawruble13:
Basic rtest.xml added in #197 .

PR #204, opened 2021-07-29
Update CMakeLists.txt

YvanMokwinski:
missing files on post build

### DIFF
--- a/clients/tests/CMakeLists.txt
+++ b/clients/tests/CMakeLists.txt
@@ -266,8 +266,8 @@ if (WIN32)
   file( GLOB third_party_dlls
     LIST_DIRECTORIES OFF
     CONFIGURE_DEPENDS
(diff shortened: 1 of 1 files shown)


TorreZuk (commented):
The start of rocblas toolchain-windows.cmake we copy it from env:
if (DEFINED ENV{HIP_DIR})
  file(TO_CMAKE_PATH "$ENV{HIP_DIR}" HIP_DIR)
  set(rocm_bin "${HIP_DIR}/bin")
else()
  set(HIP_DIR "C:/hip")
  set(rocm_bin "C:/hip/bin")
endif()
Maybe you don't have this?

YvanMokwinski:
I don't but is it a must have? the main issue was the missing ENV

TorreZuk:
Only if you check HIP_DIR as a variable elsewhere, or use rocm_bin

PR #207, opened 2021-08-05
SWDEV-296233

jsandham:
Add missing algorithms for spmv and spmm


ntrost57 (approved):

YvanMokwinski (approved):
- 0,1,2,3,4,5,6 ... is redundant. There's already a mechanism for numbering.
- shall we test?

YvanMokwinski (approved):
LGTM

jsandham:
> * 0,1,2,3,4,5,6 ... is redundant. There's already a mechanism for numbering.
> * shall we test?

I assume you are referring to the enums? I have a very slight preference for the explicit numbering as I think it makes it more clear that each value has an associated number and that in particular the order that the enums appear matters. That being said I dont really care that strongly. I assume @YvanMokwinski  you want to get rid of them? If @ntrost57 you also agree then thats a consensus and I will remove the enum numbering on all the enums. Thoughts?

YvanMokwinski:
No I'm fine with explicit numbering. I was just raising the comment as I've seen it many times. The question is : are values important? maybe they are not, maybe they are not but the order is still important to keep a certain 'compatibility'.

jsandham:
> No I'm fine with explicit numbering. I was just raising the comment as I've seen it many times. The question is : are values important? maybe they are not, maybe they are not but the order is still important to keep a certain 'compatibility'.

If I am understanding your question, I think the values are important but the order isn't I don't think. For example a user might do something like;

status = HIPSPARSE_STATUS_SUCCESS;
if(status == 1)
{
    //do something
}

The ordering I don't think is important as we always access them using routines like: hipErrorToHIPSPARSEStatus(hipError_t status) which I think makes the order in which they appear not matter. As long as the values are the same for the user then the order they appear in the enum I don't think matters.

YvanMokwinski:
> > No I'm fine with explicit numbering. I was just raising the comment as I've seen it many times. The question is : are values important? maybe they are not, maybe they are not but the order is still important to keep a certain 'compatibility'.
> 
> If I am understanding your question, I think the values are important but the order isn't I don't think. For example a user might do something like;
> 
> status = HIPSPARSE_STATUS_SUCCESS;
> if(status == 1)
> {
> //do something
> }
> 
> The ordering I don't think is important as we always access them using routines like: hipErrorToHIPSPARSEStatus(hipError_t status) which I think makes the order in which they appear not matter. As long as the values are the same for the user then the order they appear in the enum I don't think matters.
> 
> So I guess what I meant in my previous was that the value matters but not the order (although it probably makes sense to maintain the ordering increasing with value?)

I understand your point but as user, doing if status==1 is not the greatest thing to write.
Anyway if we want to keep compatibility then values can't change. Either the numbering is explicit and then order doesn't matter (you can change the order but I don't see the need for it), or the numbering is implicit and then defined by the order.

On my question: "shall we test?", it means that on the fact that values can't change, to ensure compatibility maybe we can write some tests (not now) to enforce values.

ASSERT_EQ( ((int)a), 1);
ASSERT_EQ( ((int)b), 2);
etc ....

this way if something is changed accidentally we can catch it right away, avoiding the need of understanding what's been broken in future compatibility bugs.

jsandham:
> > > No I'm fine with explicit numbering. I was just raising the comment as I've seen it many times. The question is : are values important? maybe they are not, maybe they are not but the order is still important to keep a certain 'compatibility'.
> > 
> > 
> > If I am understanding your question, I think the values are important but the order isn't I don't think. For example a user might do something like;
> > status = HIPSPARSE_STATUS_SUCCESS;
> > if(status == 1)
> > {
> > //do something
> > }
> > The ordering I don't think is important as we always access them using routines like: hipErrorToHIPSPARSEStatus(hipError_t status) which I think makes the order in which they appear not matter. As long as the values are the same for the user then the order they appear in the enum I don't think matters.
> > So I guess what I meant in my previous was that the value matters but not the order (although it probably makes sense to maintain the ordering increasing with value?)
> 
> I understand your point but as user, doing if status==1 is not the greatest thing to write.
> Anyway if we want to keep compatibility then values can't change. Either the numbering is explicit and then order doesn't matter (you can change the order but I don't see the need for it), or the numbering is implicit and then defined by the order.
> 
> On my question: "shall we test?", it means that on the fact that values can't change, to ensure compatibility maybe we can write some tests (not now) to enforce value correctness.
> 
> ASSERT_EQ( ((int)a), 1);
> ASSERT_EQ( ((int)b), 2);
> etc ....
> 
> this way if something is changed accidentally we can catch it right away, avoiding the need of understanding what's been broken in future compatibility bugs.

Sure we can test.

PR #212, opened 2021-08-12
explicitly state hip module/config dependency

ntrost57:


haampie:
Thanks, that should solve my issue!

jsandham (approved):

PR #214, opened 2021-08-19
SWDEV-294844

jsandham:
Get hipsparse compiling and all tests passing when using cuda 11.4 backend

### DIFF
--- a/clients/tests/test_spmm_bell.cpp
+++ b/clients/tests/test_spmm_bell.cpp
@@ -25,6 +25,7 @@
 
 #include <hipsparse.h>
 
+#if(!defined(CUDART_VERSION))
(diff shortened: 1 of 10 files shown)


clients/tests/test_spmm_bell.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <hipsparse.h>
 
+#if(!defined(CUDART_VERSION))
```
jsandham, line 28: spmm for bell format is only supported with CUDA_R_16F for matrices A and B when using compute capability 7.0. On the newest GPU's CUDA_R_32F becomes supported. Therefore turning of cuda tests.

ntrost57 (approved):

PR #215, opened 2021-09-01
Fix csrmm

jsandham:
Fixes were made to csrmm in rocsparse but these changes have not yet been updated in hipsparse


clients/include/testing_csrmm.hpp
```diff
@@ -447,45 +431,43 @@ hipsparseStatus_t testing_csrmm(Arguments argus)
...
-    ldc     = order == HIPSPARSE_ORDER_COLUMN ? 2 * M : 2 * N;
-
-    int nrowB = order == HIPSPARSE_ORDER_COLUMN ? ldb : B_m;
-    int ncolB = order == HIPSPARSE_ORDER_COLUMN ? B_n : ldb;
-    int nrowC = order == HIPSPARSE_ORDER_COLUMN ? ldc : C_m;
-    int ncolC = order == HIPSPARSE_ORDER_COLUMN ? C_n : ldc;
-
-    int Bnnz = nrowB * ncolB;
-    int Cnnz = nrowC * ncolC;
+    int A_m = M;
```
jsandham, line 452: In csrmm, A is always mxk (unlike in spmm_csr where op(A) is mxk)

YvanMokwinski (approved):
LGTM

clients/include/testing_csrmm.hpp
YvanMokwinski: ? in spmm_csr op(A) is mxk?..

jsandham: Yes I think so. From the cusparse documentation for spmm: 

"where op(A) is a sparse matrix with dimensions m × k "

I think this is what your PR was fixing in csrmm because I mistakenly thought they both followed this but its only spmm and not csrmm?

jsandham: This is why your PR only needed to change testing_csrmm and not testing_spmm_csr right?

jsandham: From cusparse documentation for csrmm:

"A is an m×k sparse matrix that is defined in CSR..."

jsandham: If my understanding is wrong though please correct me. I assumed that is was the bug that your PR was fixing

ntrost57: I think A is always `m x k`, you are passing A to the function, not op(A).

jsandham: mmm ok the cusparse documentation is confusing. When they say "where op(A) is a sparse matrix with dimensions m × k " I assumed then that when transA = transpose, then op(A) = A^T is mxk which means that the matrix A that you pass to the routine is kxm

jsandham: Ok ill double check that we are doing spmm_csr correctly in rocsparse. Either way I think that this PR is correct as it applies the changes you made to csrmm in rocsparse, to csrmm in hipsparse.

PR #231, opened 2021-11-10
spgemm reuse

YvanMokwinski:
Implementation of spgemm reuse.


clients/tests/test_spgemmreuse_csr.cpp
```diff
@@ -0,0 +1,58 @@
+/* ************************************************************************
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
```
jsandham, line 2: 2021

library/include/hipsparse.h
```diff
@@ -9197,6 +9199,69 @@ hipsparseStatus_t hipsparseSpGEMM_copy(hipsparseHandle_t      handle,
                                        hipsparseSpGEMMDescr_t spgemmDescr);
 #endif
 
+#if(!defined(CUDART_VERSION) || CUDART_VERSION >= 11000)
```
jsandham, line 9202: I think spgemm reuse only exists in cusparse 11.3.1 and greater. This will fail to compile with cusparse backend I think

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -11466,6 +11470,125 @@ hipsparseStatus_t hipsparseSpGEMM_copy(hipsparseHandle_t      handle,
 }
 #endif
 
+#if(!defined(CUDART_VERSION) || CUDART_VERSION >= 11000)
+HIPSPARSE_EXPORT
+hipsparseStatus_t hipsparseSpGEMMreuse_workEstimation(hipsparseHandle_t      handle,
```
jsandham, line 11475: I think spgemm reuse only exists in cusparse 11.3.1 and greater. This will fail to compile with cusparse backend I think

clients/include/testing_spgemmreuse_csr.hpp
```diff
@@ -0,0 +1,600 @@
...
+
+#include <hipsparse.h>
+#include <string>
+#include <typeinfo>
+
+using namespace hipsparse_test;
+
+void testing_spgemmreuse_csr_bad_arg(void)
+{
+#if(!defined(CUDART_VERSION) || CUDART_VERSION >= 11000)
```
jsandham, line 40: I think spgemm reuse only exists in cusparse 11.3.1 and greater. This will fail to compile with cusparse backend I think

library/src/nvcc_detail/hipsparse.cpp
YvanMokwinski: confirmed! changing this right away

clients/include/testing_spgemmreuse_csr.hpp
```diff
@@ -0,0 +1,600 @@
...
+    T* dcsr_val_C_1     = (T*)dcsr_val_C_1_managed.get();
+    J* dcsr_col_ind_C_2 = (J*)dcsr_col_ind_C_2_managed.get();
+    T* dcsr_val_C_2     = (T*)dcsr_val_C_2_managed.get();
+    if(!dcsr_col_ind_C_1 || !dcsr_val_C_1 || !dcsr_col_ind_C_2 || !dcsr_val_C_2)
+    {
+        verify_hipsparse_status_success(
+            HIPSPARSE_STATUS_ALLOC_FAILED,
+            "!dcsr_col_ind_C_1 || !dcsr_val_C_1 || !dcsr_col_ind_C_2 || !dcsr_val_C_2");
+
+        return HIPSPARSE_STATUS_ALLOC_FAILED;
```
jsandham, line 459: Note if we return here then we are not free'ing all of your hipMalloc's (specifically buffer 3 and 4)

library/src/nvcc_detail/hipsparse.cpp
```diff
@@ -11466,6 +11470,124 @@ hipsparseStatus_t hipsparseSpGEMM_copy(hipsparseHandle_t      handle,
...
+                                                                 (cusparseSpMatDescr_t)matC,
+                                                                 hipSpGEMMAlgToCudaSpGEMMAlg(alg),
+                                                                 (cusparseSpGEMMDescr_t)spgemmDescr,
+                                                                 bufferSize5,
+                                                                 externalBuffer5));
+}
+
+#endif
+
 #if(CUDART_VERSION >= 11020)
```
jsandham, line 11591: I just noticed this as well. cusparseSDDMM was only added in cusparse 11.2.2. This passed CI as I tihnk we use cusparse 11.4 there. Do you mind changing this to 11022? If a user uses cusparse 11.2.0 or 11.2.1 I think this will cause a compilation failure

clients/include/testing_spgemmreuse_csr.hpp
YvanMokwinski: good catch, I need to wrap these buffers. any hip return would leave these unfreed anyway.

library/src/nvcc_detail/hipsparse.cpp
jsandham: Also I think non-reuse spgemm was added in 11.0.3 but currently it uses 11000 here I think so that should be updated. We should probably do a review of our defines at some point but it is tedious to do.

jsandham: The nice part is that in the rocsparse channel I have updated colins math libraries api compatibility spread sheet for when cusparse generic routines were added so we can reference that.

jsandham (approved):

ntrost57 (approved):
lgtm

clients/include/testing_spgemmreuse_csr.hpp
YvanMokwinski: wrapped it with unique ptr and reset when needed.

library/src/nvcc_detail/hipsparse.cpp
YvanMokwinski: changed to 11022

PR #239, opened 2021-11-24
Fix bug in install script

jsandham:
If exporting CXX=hipcc and then calling the install script, this will result in googletest being built with CXX compiler (in this case hipcc) but hipsparse will be built with whatever c++ compiler cmake finds (the install script changes what CXX is). This means then that if googletest is built with hipcc it will generate a position dependent static library for googletest. Then when running the install script cmake will likely pick up either g++ or c++ compiler to build hipsparse. This is fine except that these compilers will build the hipsparse-test executable using position independent code. SInce our googletest static library is not position independent then this causes an error saying we must compile googtest with fPIC flag when linking. This proposed change makes it so that the compiler defaults to whatever is set by CXX so that everything is built with the same compiler. Additionally it adds the option to explicitly set the compiler used (useful for example if you are not exporting CXX variables outside the install script but still want to choose what c++ compiler you use)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -324,6 +324,12 @@ while true; do
   esac
 done
 
+# Note 'compiler' variable not changed after this point. This ensures that everything is built with the same compiler
+# provided by the user through either setting CXX externally (in which case compiler=${CXX} followed by CXX=${compiler})
+# or through using --compiler option. This is important so that googletest and hipsparse are built with the same 
+# compiler to ensure settings like position independent code is consistent when linking.
+CXX=${compiler}
+
 if [[ "${build_relocatable}" == true ]]; then
     if ! [ -z ${ROCM_PATH+x} ]; then
         rocm_path=${ROCM_PATH}
--- a/install.sh
+++ b/install.sh
@@ -19,6 +19,7 @@ function display_help()
   echo "    [-g|--debug] -DCMAKE_BUILD_TYPE=Debug (default is =Release)"
   echo "    [-k|--relwithdebinfo] -DCMAKE_BUILD_TYPE=RelWithDebInfo"
   echo "    [--codecoverage] build with code coverage profiling enabled"
+  echo "    [--compiler] specify host compiler"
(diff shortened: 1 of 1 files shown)


TorreZuk:
Curious @daineAMD if you have seen this on hipBLAS?   We should probably have a standard default compiler for hip libs if CXX env is not set.   Not sure if we build googletest won't your cmake find it and then you could still get the mismatch?   I'm leaning toward dependencies that don't need hipcc get built without it so they are more "standard" whatever that means.   Another item for the library standardization table.

YvanMokwinski (approved):

jsandham:
> Curious @daineAMD if you have seen this on hipBLAS? We should probably have a standard default compiler for hip libs if CXX env is not set. Not sure if we build googletest won't your cmake find it and then you could still get the mismatch? I'm leaning toward dependencies that don't need hipcc get built without it so they are more "standard" whatever that means. Another item for the library standardization table.

In the case of hipsparse we dont need hipcc at all as we dont have kernel code (it is important though that it is still an option as I was told Gibraltar does not have gcc available and they are expecting to use hipcc). For example when I build hipsparse from source, I don't export CXX first. Instead I just run the install script. Doing so results in cmake using /usr/bin/c++ compiler for both googletest and hipsparse. My thinking here is that using one compiler for googletest and another for hipsparse can risk running  into bugs because of how the different compilers were configured. For example g++/c++ I think have PIE code turned on by default, but I think one could build these compilers so that it is turned off. What I am trying to achieve in this PR is have whatever compiler the user has chosen be the compiler used for both googletest and hipsparse in the hopes it avoids any possible configuration issues when linking. Not sure if this is the best direction or not though so open to suggestions. I suspect hipblas has not had a bug filed about this possibly because QA is using your --compiler flag to set hipcc to be the compiler used for both googletest and hipblas so there is no inconsistency.

ntrost57 (approved):

ntrost57:
Good catch :+1:

PR #241, opened 2021-12-08
Cpp check

YvanMokwinski:
cppcheck clean up
cd build/release
and use ../../scripts/hipsparse-cppcheck.py

### DIFF
--- a/scripts/hipsparse-cppcheck.py
+++ b/scripts/hipsparse-cppcheck.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+# ########################################################################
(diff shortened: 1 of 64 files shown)


YvanMokwinski:
if you pass -Wall on compiler options there is a lot to do ...

scripts/hipsparse-cppcheck.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
jsandham, line 1: This is nice to have as a script

jsandham (approved):

ntrost57 (approved):

PR #242, opened 2022-01-18
Fix gtsv null parameter

jsandham:
Allow passing nullptrs to hipsparse gtsv/gpsv buffer size routines to make them have behaviour consistent with cusparse.


ntrost57 (approved):
That's somewhat stupid... but ok

YvanMokwinski (approved):

PR #246, opened 2022-02-01
Spmv fix

jsandham:
Ensure behaviour in hipsparse is consistent regardless of which backend is used.

### DIFF
--- a/library/src/hcc_detail/hipsparse.cpp
+++ b/library/src/hcc_detail/hipsparse.cpp
@@ -13381,8 +13381,6 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
         return rocSPARSEStatusToHIPStatus(rocsparse_status_invalid_pointer);
     }
 
-    std::cout << "hipsparseSpMV_bufferSize called" << std::endl;
-
     *bufferSize = 4;
     return HIPSPARSE_STATUS_SUCCESS;
 }
@@ -13398,7 +13396,6 @@ hipsparseStatus_t hipsparseSpMV_preprocess(hipsparseHandle_t           handle,
                                            hipsparseSpMVAlg_t          alg,
                                            void*                       externalBuffer)
 {
-    std::cout << "hipsparseSpMV_preprocess called" << std::endl;
     size_t bufferSize;
     return rocSPARSEStatusToHIPStatus(rocsparse_spmv((rocsparse_handle)handle,
                                                      hipOperationToHCCOperation(opA),
--- a/library/src/hcc_detail/hipsparse.cpp
+++ b/library/src/hcc_detail/hipsparse.cpp
@@ -13370,6 +13370,21 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
                                            hipsparseSpMVAlg_t          alg,
                                            size_t*                     bufferSize)
 {
+    if(handle == nullptr)
+    {
+        return rocSPARSEStatusToHIPStatus(rocsparse_status_invalid_handle);
+    }
+
+    if(matA == nullptr || 
+       vecX == nullptr || 
+       vecY == nullptr || 
+       alpha == nullptr || 
+       beta == nullptr || 
+       bufferSize == nullptr)
+    {
+        return rocSPARSEStatusToHIPStatus(rocsparse_status_invalid_pointer);
+    }
+
     *bufferSize = 0;
     return HIPSPARSE_STATUS_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


library/src/hcc_detail/hipsparse.cpp
```diff
@@ -13370,17 +13370,8 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
...
-                                                     alpha,
-                                                     (const rocsparse_spmat_descr)matA,
-                                                     (const rocsparse_dnvec_descr)vecX,
-                                                     beta,
-                                                     (const rocsparse_dnvec_descr)vecY,
-                                                     hipDataTypeToHCCDataType(computeType),
-                                                     hipSpMVAlgToHCCSpMVAlg(alg),
-                                                     bufferSize,
-                                                     nullptr));
+    *bufferSize = 0;
```
jsandham, line 13373: Need to add checks for invalid parameters like alpha, trans, etc.

jsandham: Otherwise this fails our bad args tests

jsandham: Fixed

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -13370,6 +13370,36 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
...
+                                           const void*                 alpha,
+                                           const hipsparseSpMatDescr_t matA,
+                                           const hipsparseDnVecDescr_t vecX,
+                                           const void*                 beta,
+                                           const hipsparseDnVecDescr_t vecY,
+                                           hipDataType                 computeType,
+                                           hipsparseSpMVAlg_t          alg,
+                                           void*                       externalBuffer)
+{
+    std::cout << "hipsparseSpMV_preprocess called" << std::endl;
```
ntrost57, line 13401: I think you have to remove this line ;)

ntrost57 (approved):

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -13370,6 +13370,36 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
...
+        return rocSPARSEStatusToHIPStatus(rocsparse_status_invalid_handle);
+    }
+
+    if(matA == nullptr || vecX == nullptr || vecY == nullptr || alpha == nullptr || beta == nullptr
+       || bufferSize == nullptr)
+    {
+        return rocSPARSEStatusToHIPStatus(rocsparse_status_invalid_pointer);
+    }
+
+    std::cout << "hipsparseSpMV_bufferSize called" << std::endl;
```
YvanMokwinski: left over

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -13370,6 +13370,36 @@ hipsparseStatus_t hipsparseSpMV_bufferSize(hipsparseHandle_t           handle,
...
+
+    *bufferSize = 4;
```
YvanMokwinski, line 13384: 0?

library/src/hcc_detail/hipsparse.cpp
```diff
@@ -13714,7 +13744,6 @@ hipsparseStatus_t hipsparseSpGEMM_copy(hipsparseHandle_t      handle,
     return status;
 }
 
-HIPSPARSE_EXPORT
```
YvanMokwinski, line 13717: why do we remove this keyword?

YvanMokwinski (approved):

library/src/hcc_detail/hipsparse.cpp
jsandham: Its not needed in the .cpp file. Only required in the .h file. I removed the few instances of it in this file as I assume they were probably there from accidentally copying the signature from the .h file to the .cpp file

library/src/hcc_detail/hipsparse.cpp
jsandham: Fixed

library/src/hcc_detail/hipsparse.cpp
jsandham: Fixed

library/src/hcc_detail/hipsparse.cpp
jsandham: Currently cannot be zero because that would mean doing:

void* temp_buffer = nullptr;
hipMalloc(&temp_buffer, buffer_size);

would still result in temp_buffer being nullptr and in the rocsparse code temp_buffer = nullptr means (when passed to rocsparse_spmv() ) calculate the buffer size. So we need to do our refactor to spmv in rocsparse at the next major release to add the 'stage' enum before we can set the buffer size to 0.

YvanMokwinski: Thanks ! again ill-specified spmv ....

PR #248, opened 2022-02-14
Set fortran samples to depend on hipsparse_fortran

lawruble13:
Previously the hipSPARSE samples used hipsparse_fortran, but didn't declare a CMake dependency on it. If ccache was enabled, the C library would build faster than the fortran module, and the fortran samples would start building. Declaring a CMake dependency should prevent this issue.


YvanMokwinski (approved):
- there is 2022 missing in the copyright
- would prefer something based on file extension (get_file_component) rather than the name itself, but it's fine.

ntrost57 (approved):

jsandham (approved):

PR #251, opened 2022-02-26
Fix cusparse test failures

jsandham:


clients/tests/test_gebsr2gebsc.cpp
```diff
@@ -44,8 +44,7 @@ hipsparseAction_t gebsr2gebsc_action_range[]
 hipsparseIndexBase_t gebsr2gebsc_csr_base_range[]
     = {HIPSPARSE_INDEX_BASE_ZERO, HIPSPARSE_INDEX_BASE_ONE};
 
-std::string gebsr2gebsc_bin[] = {"rma10.bin",
```
jsandham, line 47: Removed rma10 matrix as it was consuming too much memory for double precision and 16x16 BSR blocks.

ntrost57 (approved):

PR #271, opened 2022-06-16
[SWDEV-338627] Add testing_mult as operator overloading no longer works

jsandham:
Need to add testing_mult to call hipCmul when using hipComplex as operator overloading no longer works in newest hip runtime.


YvanMokwinski (approved):
LGTM

ntrost57 (approved):

PR #277, opened 2022-08-03
enum change to match cusparse

ntrost57:
Match cusparse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log for hipSPARSE
 
+## hipSPARSE 2.3.3
+### Changed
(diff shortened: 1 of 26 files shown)


YvanMokwinski (approved):
I don't understand what kind of match is needed here, in other words why do we have to rename HIPSPARSE_ORDER_COLUMN to _COL because of CUSPARSE_ORDER_COL.

ntrost57:
To match cusparse. We are supposed to have identical API for portability.

ntrost57:
Should hold this one back for next major release tho

jsandham (approved):

ntrost57:
I have changed HIPSPARSE_ORDER_COLUMN to be deprecated instead of being removed.

PR #284, opened 2022-10-27
Add more detailed warning when a test matrix file is not found

jsandham:
Adds more detailed warning when a test matrix file is not found to aid users and QA when running the test suite fails because a file was not found.

### DIFF
--- a/clients/include/utility.hpp
+++ b/clients/include/utility.hpp
@@ -732,7 +732,9 @@ int read_mtx_matrix(const char*          filename,
     FILE* f = fopen(filename, "r");
     if(!f)
     {
-        printf("Failed to open matrix file %s because it does not exist. Please download the matrix file using the install script with -c flag.", filename);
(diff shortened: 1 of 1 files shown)


YvanMokwinski (approved):
fprintf(stderr, or std::cerr ?

jsandham:
> fprintf(stderr, or std::cerr ?

We could use std err. I didnt because I was worried that QA would not see it unless they specifically look there. I generally dont use std error specifically for this reason because you dont see it unless you know to look for it. I used printf because that was what we used else where in the file.

YvanMokwinski:
> > fprintf(stderr, or std::cerr ?
> 
> We could use std err. I didnt because I was worried that QA would not see it unless they specifically look there. I generally dont use std error specifically for this reason because you dont see it unless you know to look for it. I used printf because that was what we used else where in the file.

Sorry, I'm not sure I understand.

ntrost57 (approved):

PR #301, opened 2023-01-13
Add const generic descriptors to hipsparse

Angeloo01:
resolves [SWDEV-374035](https://ontrack-internal.amd.com/browse/SWDEV-374035)

Summary of proposed changes:
 - Add const_spvec_descr, const_spmat_descr, const_dnvec_descr, const_dnmat_descr
 - Add and update corresponding auxiliary functions to accommodate new descriptors
 - create pre and post cuda 12 versions of affected auxiliary routines with cuda backend

Have yet to build and test on cuda 12 due to removed functions from cuda not yet propagated to hip. Builds and passes on cuda 11.4


clients/include/testing_const_dnmat_descr.hpp
```diff
@@ -0,0 +1,135 @@
+/* ************************************************************************
+ * Copyright (C) 2020 Advanced Micro Devices, Inc. All rights Reserved.
```
jsandham, line 2: 2023

clients/include/testing_const_dnvec_descr.hpp
```diff
@@ -0,0 +1,98 @@
+/* ************************************************************************
+ * Copyright (C) 2020 Advanced Micro Devices, Inc. All rights Reserved.
```
jsandham, line 2: 2023

jsandham (approved):

clients/include/testing_const_dnvec_descr.hpp
Angeloo01: fixed

clients/include/testing_const_dnmat_descr.hpp
Angeloo01: fixed

ntrost57:
cusparse did not deprecated any of those functions? They just replaced the non const version with const? I mean, this will break existing applications, no?

Angeloo01:
> cusparse did not deprecated any of those functions? They just replaced the non const version with const? I mean, this will break existing applications, no?

Passing in a non-const descriptor into a const descriptor parameter will work. So existing applications with the original descriptors would still work with the new functions. If that's what you mean?

ntrost57:
> > cusparse did not deprecated any of those functions? They just replaced the non const version with const? I mean, this will break existing applications, no?
> 
> Passing in a non-const descriptor into a const descriptor parameter will work. So existing applications with the original descriptors would still work with the new functions. If that's what you mean?

Yes. Guess I missed it's just a typedef.

ntrost57 (approved):

YvanMokwinski (approved):

PR #302, opened 2023-01-17
Remove calls to removed rocsparse routines

jsandham:
We removed deprecated rocsparse routines. This caused some errors because we were still calling deprecated rocsparse routines from hipsparse. This PR ensures that hipsparse is not longer calling these now removed routines.


YvanMokwinski (approved):
merge

PR #303, opened 2023-01-17
Update enum values to match cusparse

Angeloo01:
resolves SWDEV-377947

Proposed changes:
 - update hipsparse enum numeric values to match cusparse numeric values

### DIFF
--- a/library/include/hipsparse.h
+++ b/library/include/hipsparse.h
@@ -6966,8 +6966,8 @@ typedef enum
 #if(CUDART_VERSION >= 12000)
 typedef enum
 {
-    HIPSPARSE_CSR2CSC_ALG_DEFAULT = 1,
-    HIPSPARSE_CSR2CSC_ALG1        = 2
+    HIPSPARSE_CSR2CSC_ALG_DEFAULT = 0,
+    HIPSPARSE_CSR2CSC_ALG1        = 1
 } hipsparseCsr2CscAlg_t;
 #elif(CUDART_VERSION >= 10010 && CUDART_VERSION < 12000)
 typedef enum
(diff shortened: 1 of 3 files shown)


jsandham:
Note: I think this is necessary so that if a user has cusparse code like:

int index = static_cast<int>(CUSPARSE_SPMV_ALG1);

and they convert their code to HIP, they get the same integer values.

library/include/hipsparse.h
```diff
@@ -8930,11 +8930,11 @@ typedef enum
...
-    HIPSPARSE_SPMV_ALG_DEFAULT = 4,
-    HIPSPARSE_SPMV_COO_ALG1    = 5,
-    HIPSPARSE_SPMV_COO_ALG2    = 6,
-    HIPSPARSE_SPMV_CSR_ALG1    = 7,
-    HIPSPARSE_SPMV_CSR_ALG2    = 8
+    HIPSPARSE_SPMV_ALG_DEFAULT = 0,
+    HIPSPARSE_SPMV_COO_ALG1    = 1,
+    HIPSPARSE_SPMV_CSR_ALG1    = 2,
+    HIPSPARSE_SPMV_CSR_ALG2    = 3,
+    HIPSPARSE_SPMV_COO_ALG2    = 4
```
ntrost57, line 8937: I do not think this is valid?

library/include/hipsparse.h
```diff
@@ -8962,29 +8962,29 @@ typedef enum
...
-    HIPSPARSE_SPMM_CSR_ALG3         = 13
+    HIPSPARSE_SPMM_ALG_DEFAULT      = 0,
+    HIPSPARSE_SPMM_COO_ALG1         = 1,
+    HIPSPARSE_SPMM_COO_ALG2         = 2,
+    HIPSPARSE_SPMM_COO_ALG3         = 3,
+    HIPSPARSE_SPMM_COO_ALG4         = 5,
+    HIPSPARSE_SPMM_CSR_ALG1         = 4,
+    HIPSPARSE_SPMM_CSR_ALG2         = 6,
+    HIPSPARSE_SPMM_CSR_ALG3         = 12,
+    HIPSPARSE_SPMM_BLOCKED_ELL_ALG1 = 13
```
ntrost57, line 8973: same here
I think cusparse removed parts of it in their latest release and renumbered.

library/include/hipsparse.h
jsandham: At first I thought it was  typo in cusparse, but they do define the enum using duplicate integers. This apparently is valid and I guess is used to indcate when two enums map to the same algorithm.

Angeloo01: I believe this just indicates that the enum elements with the same value are equal and can be used interchangeably. For example, when used in a switch statement, only one of the elements with the same value can have a case statement. The other element that doesn't have an explicit case is mapped onto the element with the same value.

jsandham (approved):

ntrost57 (approved):

ntrost57:
@YvanMokwinski review required

YvanMokwinski (approved):

PR #330, opened 2023-05-25
Fixes host dependency issue

jsandham:
Fixes SWDEV-399670 host dependency issue in being able to compile minimal hipsparse program:

#include <hipsparse/hipsparse.h>
int main() {
return 0;
}


YvanMokwinski (approved):

cgmb (approved):
LGTM. In general, the rule is that if your public headers use another library's public headers, you should use `PUBLIC` for `target_link_libraries`. If not, then you can use `PRIVATE`.

library/CMakeLists.txt
```diff
@@ -75,9 +75,10 @@ target_include_directories(hipsparse
                                    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
 if(NOT USE_CUDA)
-  target_link_libraries(hipsparse PRIVATE roc::rocsparse hip::host)
+  target_link_libraries(hipsparse PUBLIC hip::host)
+  target_link_libraries(hipsparse PRIVATE roc::rocsparse)
 else()
-  target_compile_definitions(hipsparse PRIVATE __HIP_PLATFORM_NVIDIA__)
+  target_compile_definitions(hipsparse PUBLIC __HIP_PLATFORM_NVIDIA__)
```
jsandham, line 81: Fixes same problem with cuda backend.

cgmb: One day there will be `hip::host` defined for the CUDA backend, and this will all be so much simpler...

PR #411, opened 2023-11-17
Add missing hipsparseCscGet

jsandham:


library/include/hipsparse.h
```diff
@@ -38,6 +38,8 @@
 #ifndef HIPSPARSE_H
 #define HIPSPARSE_H
 
+#define INSTANTIATE_TEST_SUITE_P INSTANTIATE_TEST_CASE_P
```
ntrost57: Why this?

jsandham: Accidently left in when I was running on system with old gtest version. Removed it now.

ntrost57 (approved):

PR #412, opened 2023-11-21
Windows fixes

jsandham:
Fixes the following:
- Uses proper C:/hipSDK path to install to when building from source 
- Fixes bug where we were not using the correct prefix and suffix on hipsparse library for windows (was libhisparse.dll and libhipsparse.dll.a instead of the correct hipsparse.dll and hipsparse.a)


ntrost57 (approved):

YvanMokwinski (approved):
LG

PR #442, opened 2024-02-27
Re-enable static library CI

samjwu:
requires static lib jobs for deps to pass: 'rocBLAS', 'rocSPARSE', 'rocPRIM'


samjwu:
passes on rhel 8 and ubuntu 22
http://math-ci.amd.com/job/static-libraries/job/static/job/hipSPARSE/job/PR-442/5/

dropping test run for rhel 9

jsandham (approved):

PR #449, opened 2024-03-25
Fix SpGEMM (SWDEV-446446)

jsandham:
Fixes SpGEMM in hipsparse. See SWDEV-446446. In this JIRA issue I have attached a hipsparse example code (spgemm.cpp). Inside this file I have annotated in comments exactly how cusparse works. This PR has hipsparse mimic cusparse and thus the attached demo code also explains the decisions made in this PR. The advantages of this (beside getting it to work) are that:
- The closer cusparse and hipsparse operate, the less likely edge cases or the ordering of user code will cause discrepancies in behaviour.
- Previously, we called rocsparse_spgemm with the stage rocsparse_spgemm_stage_compute twice: once in hipsparseSpGEMM_compute and once in hipsparseSpGEMM_copy. With this PR we only call it once.
- Additionally, prior to this PR we called hipMalloc inside hipsparseSpGEMM_copy. Technically this is against the rules as cusparse explicitly states that the SpGEMM routines do not allocate device memory internally. With my changes we do not allocate device memory in any of the SpGEMM routines.

I can go over this PR in our Wednesday meeting.


library/src/amd_detail/hipsparse.cpp
```diff
@@ -14617,60 +14824,117 @@ hipsparseStatus_t hipsparseSpGEMM_copy(hipsparseHandle_t          handle,
...
-                         (rocsparse_const_spmat_descr)matB,
-                         beta_ptr,
-                         (rocsparse_spmat_descr)matC,
-                         (rocsparse_spmat_descr)matC,
-                         hipsparse::hipDataTypeToHCCDataType(computeType),
-                         hipsparse::hipSpGEMMAlgToHCCSpGEMMAlg(alg),
-                         rocsparse_spgemm_stage_compute,
-                         &bufferSize,
-                         buffer));
+    if(csrColIndTypeC == HIPSPARSE_INDEX_32I)
```
jsandham, line 14896: Need to add rocsparse_create_identity_permutation32 and rocsparse_create_identity_permutation64 to rocsparse.

ntrost57 (approved):

ntrost57:
Do not merge before corresponding rocsparse PR is in

YvanMokwinski (approved):

PR #459, opened 2024-04-11
Fix SpSM API to match cusparse (SWDEV-456312)

jsandham:
Currently our hipsparseSpSM_solve API takes an externel buffer. This API differs from cusparseSpSM_solve which does not take this external buffer directly but instead stores the pointer to the buffer inside the hipsparseSpSMDescr. See reproducer attached to SWDEV-456312.

Edit:
For now to avoid changes to the public API until the next major release, I have left the public API alone but have fixed the internals. This should make it easy when we finally correct the issue in the future.


ntrost57:
Breaking change, need to postpone for next major

jsandham:
> Breaking change, need to postpone for next major

Is it considered a breaking change or a bug? Currently, hipsparseSpSM_solve does not match the cusparse API and thus any user who is trying to convert their cusparse code to hipsparse must change how they call the functions. I would assume this would be considered a bug as this goes against the rules on how hipsparse should work?

jsandham:
@ntrost57 I have reverted the change to the public API so that this can be merged now (I have retained the fixes to the internals)

ntrost57 (approved):

PR #461, opened 2024-04-18
Reorganize hipsparse header

jsandham:
Reorganize hipsparse header to it is more manageable.
- Separates functions into their own header. This makes it easier to navigate and find functions and makes adding documentation and deprecation warnings more manageable. Follows similar to what we did in rocsparse.


library/include/conversion/hipsparse_coo2csr.h
```diff
@@ -0,0 +1,56 @@
...
+hipsparseStatus_t hipsparseXcoo2csr(hipsparseHandle_t    handle,
+                                    const int*           cooRowInd,
+                                    int                  nnz,
+                                    int                  m,
+                                    int*                 csrRowPtr,
+                                    hipsparseIndexBase_t idxBase);
+
+#ifdef __cplusplus
+}
+#endif
```
cgmb: You can get weird behaviour if you're missing a newline at the end of a header file. I would recommend fixing them with

    find . -name '*.h*' -exec sed -i '$a\' {} +

library/CMakeLists.txt
```diff
@@ -35,8 +35,7 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse-version.h.in"
                "${PROJECT_BINARY_DIR}/include/hipsparse/hipsparse-version.h")
 
 # Copy Header files to build directory
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse.h"
-               "${PROJECT_BINARY_DIR}/include/hipsparse/hipsparse.h" COPYONLY)
+execute_process(COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/include" "${PROJECT_BINARY_DIR}/include/hipsparse")
```
cgmb, line 38: `execute_process` doesn't put any information into the generated Makefile about what input files affect what output files. I think this may break your incremental builds, as your headers will not get recopied when you change them.

library/include/conversion/hipsparse_bsr2csr.h
```diff
@@ -0,0 +1,105 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * ************************************************************************ */
+#pragma once
```
cgmb: I would recommend against using `#pragma once` in public headers, particularly if you have split your headers into a large number of files like this. GCC uses a linked-list to remember which files are guarded by `#pragma once`, which will make `#include <hipsparse/hipsparse.h>` an $O(n^2)$ operation, where $n$ is the number of hipsparse headers. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58770.

A few years ago, I wrote a utility that can convert pragma once to include guards or vice versa. You can use it to switch:

```bash
python3 -m venv venv3
. venv3/bin/activate
pip install guardonce
once2guard -r library/include -p 'path -2 | upper | prepend HIPSPARSE_' -l -s '#endif /* % */'
```

That will convert all the public headers. The change for this file would be:
```patch
diff --git a/library/include/conversion/hipsparse_bsr2csr.h b/library/include/conversion/hipsparse_bsr2csr.h
index 5b8422f..7a79e5d 100644
--- a/library/include/conversion/hipsparse_bsr2csr.h
+++ b/library/include/conversion/hipsparse_bsr2csr.h
@@ -21,7 +21,8 @@
  * THE SOFTWARE.
  *
  * ************************************************************************ */
-#pragma once
+#ifndef HIPSPARSE_CONVERSION_HIPSPARSE_BSR2CSR_H
+#define HIPSPARSE_CONVERSION_HIPSPARSE_BSR2CSR_H
 
 #ifdef __cplusplus
 extern "C" {
@@ -102,4 +103,6 @@ hipsparseStatus_t hipsparseZbsr2csr(hipsparseHandle_t         handle,
 
 #ifdef __cplusplus
 }
-#endif
\ No newline at end of file
+#endif
+
+#endif /* HIPSPARSE_CONVERSION_HIPSPARSE_BSR2CSR_H */
\ No newline at end of file
```

You can tweak [the pattern](https://github.com/cgmb/guardonce/blob/master/docs/PatternLanguage.md) as you see fit.

cgmb (commented):
I have no strong opinions on the end goal, but there's a few things to consider changing in the implementation details.

library/include/conversion/hipsparse_coo2csr.h
jsandham: Thanks. Fixed.

library/include/conversion/hipsparse_bsr2csr.h
jsandham: Thanks. Fixed.

cgmb: Ironically, the command I gave you doesn't include a newline at the end of the file. I suppose that should have been `'#endif /* % */\n'`. You'll need to repeat that find/sed fix.

ntrost57:
We will skip this on 6.2

PR #462, opened 2024-04-19
Add SpGEMM estimate memory routine

jsandham:
Add missing SpGEMM estimate memory routine that was added to cuSPARSE 12.0.1


library/src/amd_detail/hipsparse.cpp
```diff
@@ -14624,6 +14624,53 @@ hipsparseStatus_t hipsparseSpGEMM_workEstimation(hipsparseHandle_t          hand
...
+{
+    if(handle == nullptr || alpha == nullptr || beta == nullptr || matA == nullptr
+       || matB == nullptr || matC == nullptr || bufferSize3 == nullptr)
+    {
+        return HIPSPARSE_STATUS_INVALID_VALUE;
+    }
+
+    if(externalBuffer3 == nullptr)
+    {
+        *bufferSize3 = 4;
```
ntrost57: Why not put 0?

jsandham: If I put 0 then the externalBuffer3 will remain nullptr as nothing will be allocated and therefore we will never enter the else branch.

ntrost57: mmhm I see. I do not really like that.. but I guess we probably have to live with it.

ntrost57 (dismissed):

ammallya:
Imported to ROCm/rocm-libraries

PR #464, opened 2024-04-26
Fix windows failure caused by writing outside memory bounds

jsandham:
Fix windows failure caused by writing outside memory bounds (specifically when k==0 and m>0 and n>0). Solution here is to not test invalid arguments inside the valid arguments section of the testing. I have instead moved the testing of invalid m, n, and k to the bad args section of the testing (just like we did in rocsparse).


clients/include/testing_csrmm.hpp
```diff
@@ -335,65 +393,6 @@ hipsparseStatus_t testing_csrmm(Arguments argus)
     }
     int nnz = M * scale * K;
 
-    // Argument sanity check before allocating invalid memory
-    if(M <= 0 || N <= 0 || K <= 0)
```
jsandham, line 339: When k == 0 and M and N > 0 we enter here (even though this combination of arguments is in fact valid). We then call the routine using arrays that are not properly allocated (we use safe_size instead of the size they should actually be). Inside rocsparse, when beta is non-zero (as is the case here) we still need to scale C = beta * C (C is mxn so unaffected by k being 0) and this is where we write out of bounds. Solution like in rocsparse is to not do this kind of invalid arguments checks in the valid args section of the testing. This simplifies the code and fixes the issue with no loss in test coverage.

ntrost57 (approved):

PR #524, opened 2024-09-18
Update precheckincuda job settings

samjwu:
Get property and job name list from settings in XML file for runs.
```
jobNameList = auxiliary.appendJobNameList(jobNameList, 'hipSPARSE')
```

Still updates the default hardcoded OS/GPU label (ubuntu20-cuda11 to ubuntu22-cuda12) for clarity, but this is to be a fallback
```
runCI(['ubuntu22-cuda12':['anycuda']], urlJobName)
```

### DIFF
--- a/.jenkins/precheckin-cuda.groovy
+++ b/.jenkins/precheckin-cuda.groovy
@@ -57,7 +57,8 @@ def runCI =
 ci: { 
     String urlJobName = auxiliary.getTopJobName(env.BUILD_URL)
 
-    def propertyList = auxiliary.appendPropertyList(propertyList)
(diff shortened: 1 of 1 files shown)


samjwu:
precheckincuda passes (http://math-ci.amd.com/job/main/job/precheckin-cuda/job/hipSPARSE/view/change-requests/job/PR-524/2/), precheckin fails due to need to build dependencies

eidenyoshida (approved):
Approved. This only affects cuda job so we can ignore the amd jobs

jsandham (approved):

PR #530, opened 2024-09-27
Update changelog for 6.3 (#525)

jsandham:


amd-jnovotny:
Very sorry that I missed this in the original PR on the develop branch, but here and in the future cuSPARSE should change to NVIDIA CUDA cuSPARSE (subsequent references can just be to CUDA cuSPARSE). Because this is a cherry pick, it's probably easier to fix that via another PR. I can look into fixing it when I change the reference to it on index.rst, but if you're making other changes to the changelog for 6.3 and want to fix it at the same time, just let me know. Thanks!

YvanMokwinski (approved):

jsandham:
> Very sorry that I missed this in the original PR on the develop branch, but here and in the future cuSPARSE should change to NVIDIA CUDA cuSPARSE (subsequent references can just be to CUDA cuSPARSE). Because this is a cherry pick, it's probably easier to fix that via another PR. I can look into fixing it when I change the reference to it on index.rst, but if you're making other changes to the changelog for 6.3 and want to fix it at the same time, just let me know. Thanks!

Ill merge this PR and put up another one that I can cherrypick into 6.3. Should the renaming of cusparse -> NVIDIA CUDA cuSPARSE be retroactive? I.e. should I change this in previous releases as well?

amd-jnovotny:
Thanks! I guess it's better to change those in the file if it's not too cumbersome to do so, but I don't think it needs to necessarily be merged back to earlier branches. It's more of a best practice to keep doing going forward.

Issue #555, opened 2024-12-08
[Issue]: build is with libc++ fails with `fatal error: 'experimental/filesystem' file not found`


AngryLoki:
### Problem Description

Hi,

The following code fails when libc++ is used as a replacement of libstdc++:
https://github.com/ROCm/hipSPARSE/blob/54071d80f509fb62d05d371cfaefa05079451e7f/clients/common/utility.cpp#L37-L40

The problem is on hipSPARSE side. Libc++ defines `__cpp_lib_filesystem` [only if `-std=c++17` or later](https://github.com/llvm/llvm-project/blob/9eb4056144a9d2e940272ba9e9675acaa3fbb354/libcxx/include/version#L307) compiler flag is specified following https://en.cppreference.com/w/cpp/feature_test and does not implement `experimental/filesystem` (it implements normal `filesystem`). Please, consider setting `set(CMAKE_CXX_STANDARD 17)` in https://github.com/ROCm/hipSPARSE/blob/b1e2366a99e75e4ee2eee855732fdb2d5d934a39/CMakeLists.txt#L70
and
https://github.com/ROCm/hipSPARSE/blob/b1e2366a99e75e4ee2eee855732fdb2d5d934a39/clients/CMakeLists.txt#L59

### Operating System

Gentoo

### CPU

-

### GPU

-

### ROCm Version

ROCm 6.3.0

ppanchad-amd:
Hi @AngryLoki. Internal ticket has been created to investigate your issue. Thanks!

AngryLoki:
Just a small update: setting `CMAKE_CXX_STANDARD 17` is not enough, as build fails with libstdc++ too:

https://github.com/ROCm/hipSPARSE/blob/b1e2366a99e75e4ee2eee855732fdb2d5d934a39/clients/common/utility.cpp#L37-L46

1) to use `__cpp_lib_filesystem` you need to `#include <version>` (or hope that it is included transitively, which is not for libstdc++)
2) when it falls back to experimental::filesystem (due to incorrect use of `__cpp_lib_filesystem`), adding `filesystem` to `std` namespace conflicts with namespace of <chrono> - see https://godbolt.org/z/qqEodn4xP for minimal reproducer.

Overall, additionally to `CMAKE_CXX_STANDARD 17` fix looks like this:
```diff
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -34,16 +34,22 @@
[...]

PR #568, opened 2025-01-24
Add c++17 to hipsparse and deprecate c++14

jsandham:
Summary of proposed changes:

- Use C++17 by default in hipSPARSE
- This is consistent with other math libraries
- Solves https://github.com/ROCm/hipSPARSE/issues/555

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Set CXX standard
 if (CMAKE_CXX_STANDARD EQUAL 14)
-  message(WARNING "C++14 will be deprecated in the next major release")
+  message( DEPRECATION "Builds using the C++14 standard will no longer be supported in the next major release" )
 elseif(NOT CMAKE_CXX_STANDARD EQUAL 17)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -66,11 +66,20 @@ endif( )
...
+if (NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Set CXX standard
+if (CMAKE_CXX_STANDARD EQUAL 14)
+  message(WARNING "C++14 will be deprecated in the next major release")
+elseif(NOT CMAKE_CXX_STANDARD EQUAL 17)
```
TorreZuk: I think your intent is something more like this:
message( DEPRECATION "Builds using the C++14 standard will no longer be supported in the next major release" )
You are deprecating support for C++14 with this PR, not in the future

jsandham: Good point. Fixed.

ntrost57 (approved):

PR #583, opened 2025-03-23
Refactor level1 routines to be in separate header files

jsandham:


library/CMakeLists.txt
```diff
@@ -38,6 +38,20 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse-version.h.in"
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse.h"
                "${PROJECT_BINARY_DIR}/include/hipsparse/hipsparse.h" COPYONLY)
 
+set(HEADER_FILES 
+  level1/hipsparse_axpyi.h
```
YvanMokwinski: Do you think we should keep the same folder hierarchy from rocSPARSE? include/internal/level1?

library/include/level1/hipsparse_axpyi.h
```diff
@@ -0,0 +1,170 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * ************************************************************************ */
+#ifndef HIPSPARSE_LEVEL1_HIPSPARSE_AXPYI_H
```
YvanMokwinski: is HIPSPARSE_LEVEL1_ necessary? 
isn't it usually the name of the file in upper case?

YvanMokwinski (approved):

library/CMakeLists.txt
jsandham: Sure ill do that.

YvanMokwinski: I still don't know if internal was really a good idea. I mimicked from blas.
but library/include/hipsparse might be a good idea.

library/include/hipsparse.h
kliegeois, line 1: Shouldn't the date be 2018-2025 as the file has been modified?
Or is it only when we add lines?

kliegeois (approved):

library/CMakeLists.txt
```diff
@@ -38,6 +38,20 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse-version.h.in"
...
                "${PROJECT_BINARY_DIR}/include/hipsparse/hipsparse.h" COPYONLY)
 
+set(HEADER_FILES 
+  internal/level1/hipsparse_axpyi.h
+  internal/level1/hipsparse_dotci.h
+  internal/level1/hipsparse_doti.h
+  internal/level1/hipsparse_gthr.h
+  internal/level1/hipsparse_gthrz.h
+  internal/level1/hipsparse_roti.h
+  internal/level1/hipsparse_sctr.h)
```
kliegeois, line 48: Wouldn't be cleaner to do:

```
file(GLOB HEADER_FILES 
     "internal/level1/*.h"
)
```
?

If we merge all the PRs, we will have a significant number of lines in this file.

library/include/hipsparse.h
ntrost57: I guess once we touch it...?

ntrost57 (approved):

library/CMakeLists.txt
jsandham: I am not particularly worried about this but I can add a for loop once I start merging and resolving the merge conflicts.

library/include/hipsparse.h
jsandham: Fixed

PR #584, opened 2025-03-23
Refactor level2 routines to be in separate header files

jsandham:


YvanMokwinski (approved):

library/include/internal/level2/hipsparse_gemvi.h
```diff
@@ -0,0 +1,235 @@
...
+#ifndef HIPSPARSE_GEMVI_H
+#define HIPSPARSE_GEMVI_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if(!defined(CUDART_VERSION) || CUDART_VERSION < 12000)
+/*! \ingroup level2_module
+ *  \brief Dense matrix sparse vector multiplication
```
kliegeois, line 33: This formatting ...

library/include/internal/level2/hipsparse_hybmv.h
```diff
@@ -0,0 +1,212 @@
...
+#ifndef HIPSPARSE_HYBMV_H
+#define HIPSPARSE_HYBMV_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if(!defined(CUDART_VERSION) || CUDART_VERSION < 11000)
+/*! \ingroup level2_module
+*  \brief Sparse matrix vector multiplication using HYB storage format
```
kliegeois, line 33: is not consistent with this one. It is the first time that observed this. This does not need to be modified, I am just curious.

Do you know if clang format let the users to use both?

kliegeois (approved):

library/include/internal/level2/hipsparse_hybmv.h
jsandham: Clang format does not seem to care

ntrost57 (approved):

PR #587, opened 2025-03-23
Refactor precond routines to be in separate header files

jsandham:


library/CMakeLists.txt
```diff
@@ -38,6 +38,22 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse-version.h.in"
...
+    precond/hipsparse_bsrilu0.h
+    precond/hipsparse_csric0.h
+    precond/hipsparse_csrilu0.h
+    precond/hipsparse_gpsv_interleaved_batch.h
+    precond/hipsparse_gtsv_interleaved_batch.h
+    precond/hipsparse_gtsv_nopivot.h
+    precond/hipsparse_gtsv_strided_batch.h
+    precond/hipsparse_gtsv.h)
+
+foreach( file_i ${HEADER_FILES})
```
YvanMokwinski, line 78: this doesn't look to be the usual way of copying files

YvanMokwinski (approved):
same comment from level1

library/CMakeLists.txt
jsandham: At first I was going to use `execute_process` but I think this might be the better way. Cory mentioned that "_execute_process doesn't put any information into the generated Makefile about what input files affect what output files. I think this may break your incremental builds, as your headers will not get recopied when you change them._"

YvanMokwinski: We had a ticket for that, about rocSPARSE.
I think you were working on it.

YvanMokwinski: in hipgraph, we have library/include/hipgraph/

with


](https://github.com/ROCm/hipGRAPH-internal/blob/develop/library/CMakeLists.txt#L131C1-L137C41)

ntrost57 (approved):

kliegeois (approved):

PR #590, opened 2025-03-23
Refactor conversion routines to be in separate header files

jsandham:


YvanMokwinski (approved):
same comment from level1

ntrost57 (approved):

kliegeois (approved):

PR #591, opened 2025-03-24
Refactor types and auxiliary routines to be in separate header files

jsandham:


library/CMakeLists.txt
```diff
@@ -38,6 +38,17 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse-version.h.in"
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/hipsparse.h"
                "${PROJECT_BINARY_DIR}/include/hipsparse/hipsparse.h" COPYONLY)
 
+set(HEADER_FILES 
+    hipsparse_types.h
```
YvanMokwinski: hipsparse-types.h?

YvanMokwinski (approved):

ntrost57 (approved):

kliegeois (approved):

PR #594, opened 2025-03-26
First part of the hipSPARSE docs refresh

amd-jnovotny:
Part 1 of the hipSPARSE docs refresh reworks the index page/ToC and add the What is hipSPARSE page. It also adds a link to the samples and copy edits and reformatting of the install docs.

The rest of the documents are left for the next round, with the exception of standardizing the topic titles to sentence case as required.


jsandham (approved):

docs/index.rst
```diff
@@ -8,42 +8,45 @@
 hipSPARSE documentation
 ********************************************************************
 
-hipSPARSE exposes a common interface that provides basic linear algebra subroutines for sparse computation implemented on top of the AMD ROCm runtime and toolchains. hipSPARSE is a SPARSE marshalling library supporting both rocSPARSE and NVIDIA CUDA cuSPARSE as backends.
-It sits between the application and a `worker` SPARSE library, marshalling inputs into the backend library and marshalling results back to the application. hipSPARSE is created using the HIP programming language and optimized for AMD's latest discrete GPUs.
+hipSPARSE provides basic linear algebra subroutines
+for sparse computation. It's written in :doc:`HIP <hip:index>`, implemented on top
+of the AMD ROCm runtime and toolchains, and optimized for AMD discrete GPUs.
+hipSPARSE is a SPARSE marshalling library that supports both :doc:`rocSPARSE <rocsparse:index>` and NVIDIA CUDA cuSPARSE as backends.
```
YvanMokwinski: I might have a comment here. I feel the new version adds confusion.
I am not sure if the library is "written" in hip, it does not require a hip compiler. It only uses the hip library and is primarily a marshalling library. It exposes a common interface for supporting both rocSPARSE and cuSPARSE backends

amd-jnovotny: Okay, I'll rework this to clarify.

amd-jnovotny: @YvanMokwinski I've updated the text on the index page now.

docs/index.rst
```diff
@@ -8,42 +8,46 @@
 hipSPARSE documentation
 ********************************************************************
 
-hipSPARSE exposes a common interface that provides basic linear algebra subroutines for sparse computation implemented on top of the AMD ROCm runtime and toolchains. hipSPARSE is a SPARSE marshalling library supporting both rocSPARSE and NVIDIA CUDA cuSPARSE as backends.
-It sits between the application and a `worker` SPARSE library, marshalling inputs into the backend library and marshalling results back to the application. hipSPARSE is created using the HIP programming language and optimized for AMD's latest discrete GPUs.
+hipSPARSE provides basic linear algebra subroutines
```
YvanMokwinski: The critical concept of "common interface" is removed from the description, and I think that was good to start with that concept. hipSPARSE is a wrapper with a requirement to be 200% compatible with the cuSPARSE C-API. The API proposed by hipSPARSE has two backend implementations, depending on the architecture available (NVIDIA or AMD). The first backend uses cuSPARSE; the implementation is highly straightforward: hipFoo calls cuFOO. If you replace the prefix "hip" with "cu", you will transform the hipSPARSE API to the cuSPARSE API, and vice versa. This must be like that way; this is the "200% compatibility" requirement. The second backend is using rocSPARSE and acts as an "adapter". There is no requirement for rocSPARSE and cuSPARSE compatibility, and the hipSPARSE AMD backend implementation is to build the bridge between the two APIs, through the hipSPARSE API.
The use of the hip library in hipSPARSE is, in fact, minimal. We use hipMalloc/hipMemcpy/hipFree, but this library has no kernel. So, the hip compiler is not required.

amd-jnovotny: @YvanMokwinski : The mention of the common interface is in the third sentence, but I can move that up in the paragraph if you think that would be better. It currently says "hipSPARSE is a SPARSE marshalling library that presents a common interface for supporting both [rocSPARSE](https://rocm.docs.amd.com/projects/rocSPARSE/en/latest/index.html) and NVIDIA CUDA cuSPARSE as backends" Before, the documentation stated hipSPARSE was created with HIP, now it just says "uses HIP". Should I remove the reference to HIP completely?

There is a bit more information in the What is hipSPARSE page? Do you think any more information about the wrapper needs to be added there. We don't want to add too much detail to the main index page, but we could add more background there.

amd-jnovotny: I've pushed a change now to reorder the paragraph.

YvanMokwinski (approved):
Thank you!

ntrost57 (approved):

PR #600, opened 2025-04-05
Break up amd_detail/hipsparse.cpp into separate translation units

jsandham:
Breakup amd_detail/hipsparse.cpp into separate translation units


library/src/CMakeLists.txt
```diff
@@ -24,7 +24,91 @@
...
+      src/amd_detail/conversion/hipsparse_hyb2csr.cpp
+      src/amd_detail/conversion/hipsparse_nnz_compress.cpp
+      src/amd_detail/conversion/hipsparse_nnz.cpp
+      src/amd_detail/conversion/hipsparse_prune_csr2csr_by_percentage.cpp
+      src/amd_detail/conversion/hipsparse_prune_csr2csr.cpp
+      src/amd_detail/conversion/hipsparse_prune_dense2csr_by_percentage.cpp
+      src/amd_detail/conversion/hipsparse_prune_dense2csr.cpp
+      # Reorder
+      src/amd_detail/reorder/hipsparse_csrcolor.cpp
+      )
```
kliegeois: I am not sure if this is on purpose to list all files explicitly but there are alternatives to include all `*.cpp` files if preferred.

kliegeois (approved):
I checked the git diff between the branch and develop and I did not see any dropped line or added line except the changes related to the separation in multiple files.

ntrost57 (approved):

library/src/CMakeLists.txt
YvanMokwinski: FILE(GLOB AMD_DETAIL_SOURCE *.cpp)

YvanMokwinski: btw, it may be better to have a cmake script in amd_detail and another one in nvidia_detail.
See hipGRAPH-internal.

YvanMokwinski (approved):

PR #605, opened 2025-04-15
update the gfortran pkg dependency for almalinux

srekolam:
The PR updates the pkg requirement for gfortran . when built using almalinux and tried on other distros like rhel8, there is dependency on libgfortran4 which is not correct.


pramenku:
This PR might break SLES 15. Refer https://github.com/ROCm/hipBLAS/pull/994#discussion_r2050988827

srekolam:
@ntrost57 can you please review.

jsandham (approved):

PR #609, opened 2025-04-25
Add compilation of C program using hipSPARSE

jsandham:
Adds example_axpyi.c program to hipSPARSE so that using hipSPARSE with C programs is tested by CI


YvanMokwinski (approved):
Thanks!

kliegeois (approved):

PR #634, opened 2025-06-29
Add missing GTest includes and linking

ckastner:
The missing gtest.h causes a compile error because of GTEST_SKIP use, which also seems to be the cause of a linking issue:

```
/usr/bin/ld: CMakeFiles/hipsparse-bench.dir/hipsparse_routine.cpp.o: in function `testing_coo2csr<float>(Arguments)::{lambda()#1}::operator()() const':
./obj-x86_64-linux-gnu/clients/benchmarks/./clients/benchmarks/../include/testing_coo2csr.hpp:103:(.text+0x6fe9b): undefined reference to `testing::Message::Message()'
```

This patch resolves the issue on my end.

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -21,6 +21,8 @@
 #
 # ########################################################################
 
+find_package(GTest REQUIRED)
+
 set(HIPSPARSE_BENCHMARK_SOURCES
   client.cpp
   hipsparse_arguments_config.cpp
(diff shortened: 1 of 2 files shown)


YvanMokwinski:
@ckastner 
The benchmark part of the client does not use GTest. Since the testing_xxxx files are compiled for both the benchmark application and the test application, two implementations need to be provided for this functionality. The test application will use the GTest macros, but the benchmark will use its own, as it shouldn't depend on GTest.

The design is debatable and could/should be refactored. Unfortunately, we have higher priorities at the moment than rewriting the client framework.

YvanMokwinski:
This PR should be rejected.

ckastner:
I understand that the ideal solution would look different. However, the current solution fails to build at all on my end, aborting with the aforementioned compile error (fixed by including `gtest.h`) and later the linker error (fixed in cmake). 

I thought this double-use of the tests might be intentional so I didn't file an issue for that, just the PR with this workaround.

Would you prefer me to file only the fails-to-build issue instead, and close the PR? Perhaps there is a better solution that I didn't think of.

YvanMokwinski:
@ckastner 
Please improve the description of the problem, compiler error, linker error, environment etc ...

I only see a linking issue: "undefined reference to `testing::Message::Message()'", and not the compiler error, as mentioned in your description. I also don't know about including gtest in testing_bsrmv.hpp.

ckastner:
@YvanMokwinski Attempting to reproduce this in a pristine environment, I now see that the GTEST_SKIP in question is not part of the upstream 6.4.1 source, but of a specific patch in the Debianized version of it, introduced so that tests that require downloading data can be skipped.

I'm terribly sorry for the noise. Closing.

